PR #87, opened 2024-12-10
fix(tests): Fix logging integration test

enochtangg:
Small fix for improving logging in the integration test. Instead of dumping all consumer logs (GHA refuses to print all logs), when an error log is seen, capture 10 lines before and after the error and print to stdout.

### DIFF
--- a/python/integration_tests/test_consumer_rebalancing.py
+++ b/python/integration_tests/test_consumer_rebalancing.py
@@ -226,10 +226,6 @@ Running test with the following configuration:
         for log in consumer_error_logs:
             print(log)
 
-    # Clean up test output files
-    print(f"Cleaning up test output files in {TESTS_OUTPUT_PATH}")
-    shutil.rmtree(TESTS_OUTPUT_PATH)
-
     assert all([row[3] == 0 for row in row_count])
     assert consumers_have_data
     assert not consumer_error_logs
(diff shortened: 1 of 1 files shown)


python/integration_tests/test_consumer_rebalancing.py
```diff
@@ -203,41 +203,33 @@ def test_tasks_written_once_during_rebalancing() -> None:
...
+        print("\nTest failed! Lower than expected amount of kafka messages in sqlite")
+
+    if consumer_error_logs:
+        print("\nTest failed! Errors in consumer logs")
+        for log in consumer_error_logs:
+            print(log)
 
     # Clean up test output files
     print(f"Cleaning up test output files in {TESTS_OUTPUT_PATH}")
     shutil.rmtree(TESTS_OUTPUT_PATH)
```
john-z-yang, line 231: I think it would make sense to not delete the output files. Because it would be nice to check the output file manually since we're not dumping the logs in the terminal anymore

john-z-yang:
Also I'm seeing more of, which really sucks
```terminal
ERROR taskbroker: Task grpc_server failed: GRPC server task failed: tonic::transport::Error(Transport, Os { code: 98, kind: AddrInUse, message: "Address already in use" })
```

markstory:
I wonder if we're not releasing the ports that grpc is connected to quickly enough?

PR #91, opened 2024-12-11
feat(k8s): Add basic k8s config files

evanh:
This adds two k8s config files: one to create the storage class used for sqlite, and another to run the taskbroker service itself.

The service currently uses a specific cloud image. This should be fixed in the future to be locked to the latest build of the image, instead of a specific commit.

There are some rough instructions in the README on how to apply this k8s config to a sandbox for testing.

### DIFF
--- a/k8s/k8s_stateful_set.yaml
+++ b/k8s/k8s_stateful_set.yaml
@@ -11,6 +11,10 @@ spec:
   clusterIP: None
   selector:
     app: taskbroker
+  sessionAffinity: ClientIP
+  sessionAffinityConfig:
+    clientIP:
+      timeoutSeconds: 86400 # default is 10800
 ---
 apiVersion: apps/v1
 kind: StatefulSet
--- a/k8s/k8s_stateful_set.yaml
+++ b/k8s/k8s_stateful_set.yaml
@@ -1,0 +1,52 @@
+apiVersion: v1
+kind: Service
+metadata:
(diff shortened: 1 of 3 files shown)


k8s/k8s_stateful_set.yaml
```diff
@@ -0,0 +1,52 @@
...
+          mountPath: /opt/sqlite
+  volumeClaimTemplates:
+  - metadata:
+      name: sqlite
+    spec:
+      accessModes: [ "ReadWriteOnce" ]
+      storageClassName: taskbroker-sqlite
+      resources:
+        requests:
+          storage: 100Mi
```
markstory, line 52: We might need more storage space for the production use but this should be good enough to start with.

k8s/k8s_stateful_set.yaml
```diff
@@ -0,0 +1,52 @@
...
+  name: taskbroker-service
+  labels:
+    app: taskbroker-service
+spec:
+  ports:
+  - port: 50051
+    name: grpc
+  clusterIP: None
+  selector:
+    app: taskbroker
```
markstory, line 13: We'll also need to configure [session affinity](https://kubernetes.io/docs/reference/networking/virtual-ips/#session-affinity) for the grpc service so that workers can communicate with the same broker consistently.

markstory (approved):
Looks good enough to move forward with and iterate on.

PR #100, opened 2024-12-16
feat(tests): End-to-end integration test for basic task processing

enochtangg:
### Overview
This PR is responsible for adding an integration test that covers basic task consumption and completion with a lightweight 'worker'. To accomplish this, a task client called `TaskWorkerClient` was created to interface with the consumer gRPC server along with a "worker" called `SimpleTaskWorker`. The integration test starts N number of these workers and 1 taskbroker in separate threads. Synchronization events are used to instruct workers on when to process and shutdown. A sequence diagram can be found in the docstring of `test_task_worker_processing()`. As tasks are fetched and completed they are saved and written to a log file. Finally, the total number of fetched and completed tasks across all workers are checked with the total number of messages produced onto the topic.

Other small changes:
* Use unique directory for integration test outputs files
* Recreate the topic at the start of each integration test to ensure clean state.

### DIFF
--- a/python/integration_tests/test_task_worker_processing.py
+++ b/python/integration_tests/test_task_worker_processing.py
@@ -19,7 +19,14 @@ from python.integration_tests.helpers import (
 from python.integration_tests.worker import SimpleTaskWorker, TaskWorkerClient
 
 
-def manage_taskworker(worker_id: int, consumer_config: dict, log_file_path: str, num_messages: int, tasks_written_event: threading.Event, shutdown_event: threading.Event) -> None:
+def manage_taskworker(
+    worker_id: int,
+    consumer_config: dict,
+    log_file_path: str,
+    num_messages: int,
+    tasks_written_event: threading.Event,
+    shutdown_event: threading.Event
+) -> None:
     print(f"[taskworker_{worker_id}] Starting taskworker_{worker_id}")
     worker = SimpleTaskWorker(TaskWorkerClient(f"127.0.0.1:{consumer_config['grpc_port']}"))
     fetched_tasks = 0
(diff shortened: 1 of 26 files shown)


enochtangg:
This test is currently failing due duplicate tasks being fetched and completed by different workers. As we increase the number of workers and messages, the duplicate value also increases. I am investigating if this a a problem with the test or the taskbroker.

python/integration_tests/test_task_worker_processing.py
```diff
@@ -0,0 +1,278 @@
...
+            if next_task:
+                task = next_task
+                next_task = None
+                fetched_tasks += 1
+            else:
+                task = worker.fetch_task()
+                if task:
+                    fetched_tasks += 1
+            if not task:
+                if get_num_tasks_by_status(consumer_config, 'Pending') == 0:
```
markstory, line 59: Do we need to do this check? Once a worker has gotten an empty response, we can infer that there are no pending tasks remaining and can shutdown the worker. This could be contributing to contention in higher worker counts.

python/integration_tests/test_task_worker_processing.py
```diff
@@ -0,0 +1,278 @@
...
+                    print(f"[taskworker_{worker_id}]: No more pending tasks to retrieve, shutting down taskworker_{worker_id}")
+                    shutdown_event.set()
+                    break
+                time.sleep(1)
+                continue
+            else:
+                next_task = worker.process_task(task)
```
markstory, line 66: Having logs of which tasks were executed might help in identifying duplicate executions.

python/integration_tests/test_task_worker_processing.py
enochtangg: You're right, we don't need this check anymore. Originally, I had the workers processing messages in parallel with the consumer inserting into the inflight store. Now that we wait for all messages to be written first before processing, we don't need this check.

enochtangg:
Requires: https://github.com/getsentry/taskbroker/pull/106

john-z-yang (approved):

john-z-yang:
I think this is ready to go.
Merge from main after https://github.com/getsentry/taskbroker/pull/108, this is now a separate job in gha.
I think we should make those 2 integration tests (rebalance and worker processing) as requirement similar to `Tests (ubuntu)`.

PR #102, opened 2024-12-17
feat(kafka): Create the topics if they don't exist

evanh:
In dev situations, it is possible for the taskbroker to be started without the necessary topics existing locally. Add a function that will create the topics if they don't exist. This function can be skipped with a config setting.

This is necessary for a couple reasons: The taskbroker runs independently of sentry itself, in the same way snuba/relay run independently. That means it should be possible to start taskbroker without needing to start the sentry monolith. These types of applications are managed using the `devservices` commands (vs. `devserver`, which starts the monolith).

However, `devservices` has no mechanism to run commands on the dependent services, which is necessary in the `taskbroker` case to initialize the kafka topics.

In order to have `taskbroker` be independent of the sentry monolith, and to not have it require extra commands to bootstrap after running `devservices up`, it was necessary to create the topics manually when the service starts.

Note that this is also what both snuba and sentry do when they start up: https://github.com/getsentry/sentry/blob/0722d000e2560bec9c097c21cf790cdcf2101197/src/sentry/runner/commands/upgrade.py

https://github.com/getsentry/snuba/blob/master/snuba/cli/bootstrap.py#L85

### DIFF
--- a/config/config-sentry-dev.yaml
+++ b/config/config-sentry-dev.yaml
@@ -1,2 +1,3 @@
 upkeep_task_interval_ms: 2000
 kafka_cluster: 127.0.0.1:9092
+create_missing_topics: true
--- a/src/consumer/admin.rs
+++ b/src/consumer/admin.rs
@@ -1,0 +1,56 @@
+use anyhow::Error;
+use rdkafka::{
+    admin::{AdminClient, AdminOptions, NewTopic, TopicReplication},
+    metadata::Metadata,
+    ClientConfig,
+};
+use std::time::Duration;
+use tracing::info;
+
+pub async fn create_missing_topics(
+    kafka_client_config: ClientConfig,
+    topics: &[&str],
+    default_topic_partitions: i32,
+) -> Result<(), Error> {
+    let admin_client: AdminClient<_> = kafka_client_config.create().unwrap();
+    let mut topics_to_create = Vec::new();
+    for topic in topics {
+        // Weirdly, this call will actually create the topic if it doesn't exist. So it needs to be overridden in the next steps.
+        let topic_metadata: Metadata = admin_client
+            .inner()
+            .fetch_metadata(Some(topic), Duration::new(5, 0))?;
+
+        if !topic_metadata.topics().is_empty() {
+            for meta_topic in topic_metadata.topics() {
+                if meta_topic.error().is_some()
+                    && meta_topic.partitions().len() as i32 != default_topic_partitions
+                {
+                    info!(
+                        "Creating topic: {} with partitions: {}",
+                        meta_topic.name(),
+                        default_topic_partitions
+                    );
+                    topics_to_create.push(NewTopic::new(
+                        topic,
+                        default_topic_partitions,
+                        TopicReplication::Fixed(1),
+                    ));
+                } else {
+                    info!(
+                        "Using topic: {}, partitions: {}, error: {:?}",
+                        meta_topic.name(),
+                        meta_topic.partitions().len(),
+                        meta_topic.error()
+                    );
+                }
+            }
+        }
+    }
+    if !topics_to_create.is_empty() {
+        admin_client
+            .create_topics(&topics_to_create, &AdminOptions::new())
+            .await?;
+    }
+
+    Ok(())
+}
--- a/src/consumer/kafka.rs
+++ b/src/consumer/kafka.rs
@@ -4,14 +4,12 @@ use futures::{
     pin_mut, Stream, StreamExt,
 };
 use rdkafka::{
-    admin::{AdminClient, AdminOptions, NewTopic, TopicReplication},
     consumer::{
         stream_consumer::StreamPartitionQueue, BaseConsumer, Consumer, ConsumerContext, Rebalance,
         StreamConsumer,
     },
     error::{KafkaError, KafkaResult},
     message::{BorrowedMessage, OwnedMessage},
-    metadata::Metadata,
     ClientConfig, ClientContext, Message, Offset, TopicPartitionList,
 };
 use std::{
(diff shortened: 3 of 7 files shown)


src/consumer/kafka.rs
```diff
@@ -71,6 +82,56 @@ pub async fn start_consumer(
...
+        // Weirdly, this call will actually create the topic if it doesn't exist. So it needs to be overridden in the next steps.
+        let topic_metadata: Metadata = admin_client
+            .inner()
+            .fetch_metadata(Some(topic), Duration::new(5, 0))
+            .expect("Failed to fetch metadata");
+
+        if !topic_metadata.topics().is_empty() {
+            for meta_topic in topic_metadata.topics() {
+                if meta_topic.error().is_some()
+                    && meta_topic.partitions().len() as i32 != default_topic_partitions
```
markstory, line 102: Do we need to create the topic if `meta_topic` has more partitions than `default_topic_partitions`?

src/consumer/kafka.rs
```diff
@@ -42,11 +44,20 @@ use tracing::{debug, error, info, instrument};
...
         &BTreeSet<(String, i32)>,
     ) -> ActorHandles,
 ) -> Result<(), Error> {
+    create_missing_topics(
+        kafka_client_config.clone(),
+        topics,
+        default_topic_partitions,
+    )
+    .await
+    .expect("Failed to find/create topics");
```
markstory, line 59: Would we want to create/check for topics in production?

fpacifici:
Why do we need to create topics on the fly ? The topics to be created are already supposed to be in the schema registry and created by the sentry dev environment setup (it is possible this was never added and it should be added there).

john-z-yang (changes requested):
I think it would make more sense to do this as a `make` recipe if it's for dev only. Adding all this logic and dependencies into the consumer binary just for development does not seem worthwhile to me.

I also agree with others here that this seems to be encroaching on abilities/responsibilities that the consumer should not have.

src/consumer/kafka.rs
evanh: If the topic already exists and has partitions, then the `error` will be null and this check won't get hit.

src/consumer/kafka.rs
evanh: Why not? It only happens on startup and will no-op.

john-z-yang:
Discussed offline that this will likely be needed to unblock us for sandbox testing. We can make a ticket to remove this once devservices has scripts to setup topics.

src/consumer/kafka.rs
```diff
@@ -71,6 +82,56 @@ pub async fn start_consumer(
     .await
 }
 
+async fn create_missing_topics(
```
john-z-yang, line 85: Let's move this to another file since I don't think it belongs in the consumer directory.

src/consumer/kafka.rs
```diff
@@ -42,11 +44,20 @@ use tracing::{debug, error, info, instrument};
...
         Arc<StreamConsumer<KafkaContext>>,
         &BTreeSet<(String, i32)>,
     ) -> ActorHandles,
 ) -> Result<(), Error> {
+    create_missing_topics(
+        kafka_client_config.clone(),
+        topics,
+        default_topic_partitions,
+    )
+    .await
```
john-z-yang, line 58: We should call this function in `main`, before any of the green threads are spawned, so any workers do not try and make grpc connections.

src/consumer/kafka.rs
```diff
@@ -71,6 +82,56 @@ pub async fn start_consumer(
...
+                    );
+                }
+            }
+        }
+    }
+    if !topics_to_create.is_empty() {
+        admin_client
+            .create_topics(&topics_to_create, &AdminOptions::new())
+            .await
+            .expect("Failed to create topic");
```
john-z-yang, line 129: I see a couple of `.expect` here, it might be better to just use the `?` operator to get better error handling, especially when it is being called in main.

config/config-sentry-dev.yaml
```diff
@@ -1,2 +1,3 @@
 upkeep_task_interval_ms: 2000
 kafka_cluster: 127.0.0.1:9092
+create_missing_topics: true
```
john-z-yang, line 3:
```suggestion
create_missing_topics: true

```

src/consumer/admin.rs
```diff
@@ -0,0 +1,56 @@
...
+
+pub async fn create_missing_topics(
+    kafka_client_config: ClientConfig,
+    topics: &[&str],
+    default_topic_partitions: i32,
+) -> Result<(), Error> {
+    let admin_client: AdminClient<_> = kafka_client_config.create().unwrap();
+    let mut topics_to_create = Vec::new();
+    for topic in topics {
+        // Weirdly, this call will actually create the topic if it doesn't exist. So it needs to be overridden in the next steps.
```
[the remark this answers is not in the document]
evanh: It is very weird. What I experienced was: without any of this code, the consumer crashloops with the topic not existing. If all this function did was call `fetch_metadata`, then the crashloops would be solved.

The issue is the partitions and settings it uses in that case. I thought it would be better to explicitly create the topic with the settings we want, rather than assume whatever `fetch_metadata` does will work properly.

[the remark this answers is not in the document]
john-z-yang: Yeah that makes sense, so what if you just make the call to `create_topic` instead of fetching metadata?
Like
```rust
let res = admin_client
    .create_topics(
        ...,
        &AdminOptions::new()
     )
    .await?;
match res {
    Ok() | TopicResult::Err((_, RDKafkaErrorCode::TopicAlreadyExists)) => Ok(())
    err => err
}
```

config/config-sentry-dev.yaml
john-z-yang: I see this is resolved, do we not need a new line here?

john-z-yang (approved):
This looks fine to me, maybe we can consider removing it once it's out of the sandbox and we have mechanisms to create topics.

Another note, I feel like we made this a little more difficult because we're allowing the consumer to subscribe to multiple topics. Maybe we don't need this and we can clamp down on this API a little more.

PR #104, opened 2024-12-19
ref(kafka): Assume taskbroker subscribes to a single topic only

john-z-yang:
### Overview

- Turns out the topic is created when the metadata is fetched, not when the actual create topic call come in. This causes the `default_topic_partitions` flag to not work, as fetching metadata creates a topic with a single partition, and the subsequent `create_topics` results in a no-op. Change the behaviour to just create the topic without fetching metadata.
- I think taskbroker can only be configured to subscribe to a single topic right now, since the config file defines kafka_topic as a string, not a collection of strings. Updates our consumer/admin logic to reflect this and removes some code.


john-z-yang:
This still presents the edge case if someone has the following settings
```
create_missing_topics: true
kafka_topic: "blah_blah_blah_blah"
default_topic_partitions: 4
```

if the topic already exists, but it only has something like 2 partitions, this code does not scale up the number of partition from 2 → 4 since the topic is already there.

A little bit unexpected, but IMO it's okay because the setting is called `default_topic_partitions`, not `topic_partitions`, which implies that it is only used if the topic does not exists in the first place.

evanh (approved):

evanh:
> A little bit unexpected, but IMO it's okay because the setting is called default_topic_partitions, not topic_partitions, which implies that it is only used if the topic does not exists in the first place.

I agree with this. If we're in a situation where the setting the application has is different from what is initialized in kafka, that isn't something the application should try to figure out, that is an infrastructure concern that should be handled elsewhere.

PR #108, opened 2024-12-23
test(kafka): Refactor integration tests

john-z-yang:
### Overview

1. Separate the unit tests and integration test into separate CI checks
2. Unify make file and CI so that the CI uses the makefile as much as possible. This way, it's much easier to reproduce CI failure locally.
3. Destroys the kafka container and volume before starting an integration test, so all integration tests can start with a clean kafka container.
4. Update the integration test output directory structure from:
```
.
└── integration_tests/
    └── .test_output/
        └── # Files all go in here
```

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;to:

```
.
└── integration_tests/
    └── .test_output/
        ├── rebalance/
        │   └── # rebalance output goes here
        └── some_other_test/
            └── # output for the other test goes here
```
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; so that finding integration tests would be easier as we add more
5. Update the makefile recipe for the integration test so that they delete the test output later. This ensures that the output only gets cleaned up after the test returns with 0 exit code.
6. Increase minimum time before a consumer startup and it recieving a shutdown signal from 1 seconds to 4 seconds. For some reason, with a fresh kafka topic created using the python integration test. If a new consumer starts up and shuts down before a partition assignment, it will print a kafka broker error when shutting down. This trips the error output check in the integration test causing it to fail but no data is lost / double processed. ~~I would like to dive deeper to why this is happening and why it only happens when a topic is created in the python integration test, but I would like to unblock us first. Also, in a real prod environment, we never shut down our consumer that fast.~~ Okay I figured it out, kafka has a setting called `group.initial.rebalance.delay.ms` (doc [here](https://cwiki.apache.org/confluence/display/KAFKA/KIP-134)), it controls how long to wait to respond to a consumer `JoinGroupRequest` when a consumer group is created. Before, the unit tests already creates the task-worker consumer group, so when the integration test runs, there are no delays from the broker when the consumers are starting. But now, because we are recreating the kafka container from scratch, the broker waits when the consumers are sending the `JoinGroupRequest`, and since our integration test can send shutdown signal in 1 second, we get the timeout error because the consumer is sending a `LeaveGroupRequest` while the broker is still not responding to the `JoinGroupRequest` from when it first starts, causing the timeout:
```
librdkafka: REQTMOUT [thrd:GroupCoordinator]: GroupCoordinator/1001: Timed out LeaveGroupRequest in flight (after 5004ms, timeout #0): possibly held back by preceeding blocking JoinGroupRequest with timeout in 297034ms
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,13 +60,30 @@ jobs:
         with:
           key: ${{ github.job }}
 
-      - name: Run Cargo Tests
-        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # pin@v1
-        with:
-          command: test
-          args: --all
-
-      - name: Run Python Integration Tests
+      - name: Run unit tests
         run: |
-          export PYTEST_ADDOPTS=""
-          python -m pytest python/integration_tests -s -vv
+          make unit-test
+
+  rebalance-integration-test:
+    name: Rebalance integration test
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install protoc
+        uses: arduino/setup-protoc@v3
+
+      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # pin@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: swatinem/rust-cache@81d053bdb0871dcd3f10763c8cc60d0adc41762b # pin@v1
+        with:
+          key: ${{ github.job }}
+
+      - name: Run Rebalance Integration Test
+        run: |
+          make test-rebalace
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,4 +86,4 @@ jobs:
 
       - name: Run Rebalance Integration Test
         run: |
-          make test-rebalace
+          make test-rebalance
--- a/python/integration_tests/test_consumer_rebalancing.py
+++ b/python/integration_tests/test_consumer_rebalancing.py
@@ -13,11 +13,12 @@ import yaml
 
 from python.integration_tests.helpers import (
     TASKBROKER_BIN,
-    recreate_topic,
+    TESTS_OUTPUT_ROOT,
+    create_topic,
     send_messages_to_kafka,
 )
 
-TEST_OUTPUT_PATH = Path(__file__).parent / ".output_from_test_consumer_rebalancing"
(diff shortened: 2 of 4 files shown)


python/integration_tests/test_consumer_rebalancing.py
```diff
@@ -56,7 +56,7 @@ def test_tasks_written_once_during_rebalancing() -> None:
     consumer_path = str(TASKBROKER_BIN)
     num_consumers = 8
     num_messages = 100_000
-    num_restarts = 16
+    num_restarts = 2
```
john-z-yang, line 59: TODO: Revert this

.github/workflows/ci.yml
```diff
@@ -60,13 +60,30 @@ jobs:
...
-        uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # pin@v1
+      - name: Run unit tests
+        run: |
+          make unit-test
+
+  rebalance-integration-test:
+    name: Rebalance integration test
+    runs-on: ubuntu-latest
+
+    steps:
```
enochtangg, line 71: Do we need to do `devservices up` again in this separate job?

python/integration_tests/test_consumer_rebalancing.py
```diff
@@ -79,27 +80,18 @@ def test_tasks_written_once_during_rebalancing() -> None:
...
-            f"{topic_name} topic does not exist, creating it with {num_partitions} partitions"
-        )
-        create_topic(topic_name, num_partitions)
-    else:
-        print(
-            f"{topic_name} topic already exists, making sure it has {num_partitions} partitions"
-        )
-        update_topic_partitions(topic_name, num_partitions)
+    # Ensure topic exists and has correct number of partitions
+    create_topic(topic_name, num_partitions)
```
enochtangg, line 84: What happens if the topic already exists?

enochtangg (approved):
Nice find and thanks for cleaning all this stuff up. Just some small questions.

.github/workflows/ci.yml
john-z-yang: We don't because the makefile recipe has build and reset-kafka as prerequisites.
```makefile
test-rebalance: build reset-kafka
```
So when we do `make test-rebalance` it will automatically do `make build` and `make reset-kafka` for us (which does a devservices up & down)

python/integration_tests/test_consumer_rebalancing.py
john-z-yang: I think this will print an error message and continue, but usually we run this with `make test_rebalance`, it always clears the kafka container, because reset-kafka is a prerequisite

PR #109, opened 2024-12-24
feat(config): Allow config to control ansi encoding in logs

john-z-yang:


evanh (approved):

enochtangg (approved):

src/logging.rs
```diff
@@ -114,8 +118,9 @@ pub fn init(log_config: LoggingConfig) {
             .with_span_list(true)
             .with_file(true)
             .with_line_number(true)
+            .with_ansi(log_config.with_ansi)
```
markstory, line 121: When would we want JSON logging with escape codes?

src/logging.rs
```diff
@@ -114,8 +118,9 @@ pub fn init(log_config: LoggingConfig) {
...
             .boxed(),
-        LogFormat::Text => subscriber.compact().boxed(),
+        LogFormat::Text => subscriber.compact().with_ansi(log_config.with_ansi).boxed(),
```
markstory, line 123: Another approach would be to have `text` (with-ansi) and `plaintext` (or a similar name) that doesn't feature ansi formatting.

john-z-yang:
I gave this a little more thought and feel like this pr isn't really necessary since we already can set the logging output to JSON. So I am going to close this pr is nobody objects.

@evanh @enochtangg @markstory

PR #110, opened 2025-01-03
feat: Add a Docker container for taskworkers

evanh:
When testing in sandboxes, it's helpful to have a worker image that can be scaled up and down easily, and that can be configured with different failure modes. Add a new docker image for a simple worker that can be set to fail randomly a certain percentage of times.

The failure can be an actual task failure, or simply never returning a status for a task, causing it to "timeout".


python/integration_tests/worker.py
```diff
@@ -87,3 +89,44 @@ def process_task(self, activation: TaskActivation) -> TaskActivation | None:
             status=TASK_ACTIVATION_STATUS_COMPLETE,
             fetch_next_task=FetchNextTask(namespace=self._namespace),
         )
+
+
+class ConfigurableTaskWorker:
```
markstory, line 94: Perhaps in the future we can replace `SimpleTaskWorker` usage with this? If both timeout and failure rates are 0, they should be functionally the same.

markstory (approved):
Makes sense to me.

.github/workflows/image.yml
```diff
@@ -37,3 +37,38 @@ jobs:
...
+
+    - run: docker buildx create --driver docker-container --use
+
+    - name: build
+      run: |
+        set -euxo pipefail
+
+        if [ ${{ github.event_name }} = 'push' ]; then
+          args=(
+            --tag ghcr.io/getsentry/taskworker:latest
```
enochtangg, line 59: Small question: Should this and the other references be `ghcr.io/getsentry/taskbroker:latest` instead or are we calling it taskworker?

evanh: I'm calling this image `taskworker` to distinguish it from `taskbroker`. Open to suggestions on names though.

python/integration_tests/worker.py
evanh: Makes sense to me.

PR #112, opened 2025-01-09
feat(tests): End-to-end integration for upkeep retry

enochtangg:
### What does this test do?
This tests is responsible for checking the integrity of the retry mechanism implemented in the upkeep thread of taskbroker. An initial amount of messages is produced to kafka with a set number of retries in its retry policy. Then, the taskworkers fetch and update the task's status to retry. During an interval, the upkeep thread will collect these tasks and re-produce the task to kafka. This process continues until all tasks have been retried the specified number of times.

The test starts N number of taskworker(s) and a consumer in separate
threads. Synchronization events are use to instruct the taskworker(s) when start processing and shutdown. A shared data structured access by a mutex called TaskRetriedCounter is used to globally keep track of every task retried. Finally, this total number is validated alongside the number of times each individual task was retried.

### DIFF
--- a/python/integration_tests/test_upkeep_retry.py
+++ b/python/integration_tests/test_upkeep_retry.py
@@ -15,7 +15,7 @@ from python.integration_tests.helpers import (
     create_topic,
 )
 
-from python.integration_tests.worker import SimpleTaskWorker, TaskWorkerClient
+from python.integration_tests.worker import ConfigurableTaskWorker, SimpleTaskWorker, TaskWorkerClient
 
 
 TEST_OUTPUT_PATH = TESTS_OUTPUT_ROOT / "test_upkeep_retry"
@@ -59,9 +59,7 @@ def manage_taskworker(
     retries_per_task: int,
 ) -> None:
     print(f"[taskworker_{worker_id}] Starting taskworker_{worker_id}")
-    worker = SimpleTaskWorker(
-        TaskWorkerClient(f"127.0.0.1:{consumer_config['grpc_port']}")
-    )
+    worker = ConfigurableTaskWorker(TaskWorkerClient(f"127.0.0.1:{consumer_config['grpc_port']}"), retry_rate=1)
     retried_tasks = 0
     next_task = None
     task = None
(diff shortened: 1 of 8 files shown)


john-z-yang (approved):

python/integration_tests/test_upkeep_retry.py
```diff
@@ -0,0 +1,337 @@
...
+                task_retried_count = counter.get_total_retried()
+                if task_retried_count >= num_messages * retries_per_task:
+                    print(
+                        f"[taskworker_{worker_id}]: Total tasks retried reached: {task_retried_count}/{num_messages * retries_per_task}. Shutting down taskworker_{worker_id}"
+                    )
+                    shutdown_event.set()
+                    break
+                else:
+                    time.sleep(1)
+                    continue
```
markstory, line 96: In the event that not all tasks are processed should we have a max time we'll wait for completion?

python/integration_tests/test_upkeep_retry.py
```diff
@@ -0,0 +1,337 @@
...
+
+def check_num_tasks_written(consumer_config: dict) -> int:
+    attach_db_stmt = f"ATTACH DATABASE '{consumer_config['db_path']}' AS {consumer_config['db_name']};\n"
+    query = f"""SELECT count(*) as count FROM {consumer_config['db_name']}.inflight_taskactivations;"""
+    con = sqlite3.connect(consumer_config["db_path"])
+    cur = con.cursor()
+    cur.executescript(attach_db_stmt)
+    rows = cur.execute(query).fetchall()
+    count = rows[0][0]
+    return count
```
markstory, line 178: This function shows up in a few integration tests, perhaps we should have a module of common integration test helpers?

python/integration_tests/test_upkeep_retry.py
enochtangg: We should! Added a timeout for taskworkers.

python/integration_tests/test_upkeep_retry.py
enochtangg: Good catch, lifted the function and consumer configs to helpers.py

markstory (approved):
Looks good to me.

Issue #117, opened 2025-01-15
Rename inflight_taskactivations.deadletter_at


markstory:
The column `deadletter_at` isn't only used for deadlettering. It is used as a deadline after which tasks are removed from inflight tasks. During our meeting today names like:

- `purge_at` 
- `remove_at` 

We should replace the column and update surrounding logic.

PR #130, opened 2025-01-17
Rename deadletter_at -> remove_at

markstory:
The name deadletter_at was a bit misleading as activations past this
deadline were not always deadlettered. Based on an activation's retry
policy they could be discarded as well.

Refs https://github.com/getsentry/taskbroker/issues/117


enochtangg (approved):
🚀

PR #131, opened 2025-01-17
Replace DATETIME columns with INTEGER

### LINKED ISSUE #111: Use number type column for datetimes
We've hit a few problems with our datetime comparison logic as sqlite stores datetimes as strings and requires `datetime()` to be used on comparison operations.  If instead we stored timestamps as `real` or `integer` values we would avoid the lexicographic comparison failures that we've run into.

This will require changes to `InflightActivationStore` and the inflight activation schema. It looks like [sqlx](https://github.com/launchbadge/sqlx/blob/main/sqlx-sqlite/src/types/chrono.rs) will handle encoding/decoding `DateTime` and `Timezone` types into float and integer columns.

markstory:
SQlite doesn't have a specific storage type for datetime values and uses TEXT instead. This requires careful use of `datetime()` functions on value. By switching to INTEGER columns we won't get burned by future changes to datetime logic if we forget a function. It should also result in smaller database sizes. sqlx includes trait implementations to convert `DateTime<Utc>` into a timestamp already and from a timestamp into `DateTime<Utc>`.

I've also made `remove_at` required as we should not store an activation that cannot be purged in the future.

Fixes #111

### DIFF
--- a/migrations/0001_create_inflight_taskactivations.sql
+++ b/migrations/0001_create_inflight_taskactivations.sql
@@ -3,10 +3,10 @@ CREATE TABLE IF NOT EXISTS inflight_taskactivations (
     activation BLOB NOT NULL,
     partition INTEGER NOT NULL,
     offset BIGINTEGER NOT NULL,
-    added_at DATETIME NOT NULL,
-    remove_at DATETIME,
+    added_at INTEGER NOT NULL,
+    remove_at INTEGER NOT NULL,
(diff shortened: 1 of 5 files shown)


evanh (commented):
What was the scenario where the `remove_at` would be null? Do we need to default it to something?

markstory:
> What was the scenario where the remove_at would be null? Do we need to default it to something?

I don't think there was a real scenario, we just hadn't implemented the broker max inflight timeout.

Issue #116, opened 2025-01-15
Update taskbroker to use new protobufs


markstory:
Once protobufs are updated in #115, we will need to update Taskbroker to use the new protobufs and change logic around retries to use the new simpler `RetryState`.

PR #133, opened 2025-01-22
feat(taskbroker): Update to new taskbroker protobufs

enochtangg:
We've created a new version of taskbroker protobufs. This PR is responsible for updating all references (and removing deprecated ones) in taskbroker.  

Fixes: https://github.com/getsentry/taskbroker/issues/116


markstory (approved):
Looks good to me.

PR #134, opened 2025-01-23
ref(upkeep): Make do_upkeep return context and update tests

enochtangg:
There are two unit tests in `src/upkeep.rs` which are passing as false positives: 
* `test_retry_is_discarded_when_exhausted_with_retry`
* `test_retry_is_discarded_with_no_retry_state`

Both these tests assert that the inflight activation store only contains one pending task and assumes that the other retry task was discarded. However, this is not the case. Rather than being discarded, the task is being reproduced to the topic. In the [task activation state machine ](https://www.notion.so/sentry/Taskbroker-Taskworkers-74a84c8b1992480190c380e282f87778?pvs=4#90fe2c8a1abf4a2eaf98b4a97c055f22), there is no path for a retry task to be discarded or dlq'ed. Instead, it is follows the following flow:
1. Retry task is reproduced as pending and attempts is incremented
2. Worker picks up the pending task
3. Worker sees that attempts >= max_attempts - 1, and sets status to failed
4. Upkeep handles the failed task and discards it.

This PR is responsible for the following:
* Update do_upkeep to return `UpkeepResults`. This allows us to actually assert the correct state rather than depending on the inflight activation store
* Introduce `FailedTasksForwarder` (maybe there's a better name?) to organize discard and deadlettered tasks.
* Remove false positive tests
* Update other unit tests to assert the correct behaviour according to `UpkeepResults`

### DIFF
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -113,7 +113,6 @@ pub async fn do_upkeep(
             }
 
             ids.push(inflight.activation.id);
-            println!("retry.publish.success {:?}", ids);
         }
 
         // 3. Update retry tasks to complete
@@ -432,8 +431,8 @@ mod tests {
         assert!(store.store(batch.clone()).await.is_ok());
         let result_context = do_upkeep(config, store.clone(), producer).await;
 
-        assert_eq!(result_context.remove_at_expired, 1); // batch[2] is removed due to remove_at deadline
-        assert_eq!(result_context.discarded, 1); // batch[2] is discarded
+        assert_eq!(result_context.remove_at_expired, 1); // batch[0] is removed due to remove_at deadline
+        assert_eq!(result_context.discarded, 1); // batch[0] is discarded
         assert_eq!(result_context.completed, 2); // batch[1] and batch[2] are removed as completed
         assert_eq!(
             store
@@ -453,8 +452,28 @@ mod tests {
         );
 
         assert!(
-            store.get_by_id(&batch[0].activation.id).await.is_ok(),
-            "first task should remain"
+            store
+                .get_by_id(&batch[0].activation.id)
+                .await
+                .unwrap()
+                .is_none(),
+            "first task should be removed"
+        );
+        assert!(
+            store
+                .get_by_id(&batch[1].activation.id)
+                .await
+                .unwrap()
+                .is_none(),
+            "second task should be removed"
+        );
+        assert!(
+            store
+                .get_by_id(&batch[2].activation.id)
+                .await
+                .unwrap()
+                .is_some(),
+            "third task should be kept"
         );
     }
 
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -36,7 +36,7 @@ pub async fn upkeep(config: Arc<Config>, store: Arc<InflightActivationStore>) {
     loop {
         select! {
             _ = timer.tick() => {
-                do_upkeep(config.clone(), store.clone(), producer_arc.clone()).await;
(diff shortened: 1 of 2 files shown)


src/upkeep.rs
```diff
@@ -109,6 +113,7 @@ pub async fn do_upkeep(
             }
 
             ids.push(inflight.activation.id);
+            println!("retry.publish.success {:?}", ids);
```
markstory, line 116: We should remove this before merging.

src/upkeep.rs
```diff
@@ -480,8 +430,11 @@ mod tests {
         batch[2].remove_at = Utc.with_ymd_and_hms(2024, 11, 14, 21, 22, 23).unwrap();
 
         assert!(store.store(batch.clone()).await.is_ok());
-        do_upkeep(config, store.clone(), producer).await;
+        let result_context = do_upkeep(config, store.clone(), producer).await;
 
+        assert_eq!(result_context.remove_at_expired, 1); // batch[2] is removed due to remove_at deadline
+        assert_eq!(result_context.discarded, 1); // batch[2] is discarded
+        assert_eq!(result_context.completed, 2); // batch[1] and batch[2] are removed as completed
```
markstory, line 437: I think the comment on 477 is also incorrect. It is index 0 that would be discarded, not 2. The assertions for row presence on 493 are also not right. 

```rust
        assert!(
            store.get_by_id(&batch[0].activation.id).await.is_ok(),
            "first task should remain"
        );
```

Should be 

```rust
        assert!(
            store.get_by_id(&batch[0].activation.id).await.unwrap().is_none(),
            "first task should be removed"
        );
        assert!(
            store.get_by_id(&batch[1].activation.id).await.unwrap().is_none(),
            "second task should be removed"
        );
        assert!(
            store.get_by_id(&batch[2].activation.id).await.unwrap().is_some(),
            "third task should be kept"
        );
```

src/upkeep.rs
```diff
@@ -480,8 +430,11 @@ mod tests {
         batch[2].remove_at = Utc.with_ymd_and_hms(2024, 11, 14, 21, 22, 23).unwrap();
 
         assert!(store.store(batch.clone()).await.is_ok());
-        do_upkeep(config, store.clone(), producer).await;
+        let result_context = do_upkeep(config, store.clone(), producer).await;
 
+        assert_eq!(result_context.remove_at_expired, 1); // batch[2] is removed due to remove_at deadline
```
markstory, line 435:
```suggestion
        assert_eq!(result_context.remove_at_expired, 1); // batch[0] is removed due to remove_at deadline
```

src/upkeep.rs
```diff
@@ -480,8 +430,11 @@ mod tests {
...
+        assert_eq!(result_context.discarded, 1); // batch[2] is discarded
```
markstory, line 436:
```suggestion
        assert_eq!(result_context.discarded, 1); // batch[0] is discarded
```

src/upkeep.rs
```diff
@@ -74,7 +77,7 @@ pub async fn do_upkeep(
     config: Arc<Config>,
     store: Arc<InflightActivationStore>,
     producer: Arc<FutureProducer>,
-) {
+) -> UpkeepResults {
```
markstory, line 80: :+1: Nice, this will help us write better tests.

src/upkeep.rs
enochtangg: Oops yes

src/upkeep.rs
enochtangg: Good catch, I had the comment the other way around. Also updated the assertion for row presence.

PR #140, opened 2025-01-27
ref(tele): Try using DogStatsD exporter

john-z-yang:


src/metrics.rs
```diff
@@ -1,34 +1,26 @@
...
 pub fn init(metrics_config: MetricsConfig) {
-    let address = metrics_config.statsd_addr;
-
-    let recorder = StatsdBuilder::from(address.ip().to_string(), address.port())
-        .with_queue_size(5000)
-        .with_buffer_size(1024)
-        .build(Some("taskbroker"))
-        .expect("Could not create StatsdRecorder");
-
-    metrics::set_global_recorder(recorder).expect("Could not set global metrics recorder")
```
evanh, line 33: Do you still need to call `metrics::set_global_recorder` on the dogstatsd builder?

john-z-yang: According to their example it does not seem like it
https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-dogstatsd/examples/dogstatsd_idle_counter.rs#L9-L16

evanh (approved):

PR #141, opened 2025-01-28
feat(tests): End-to-end integration for upkeep DLQ 

enochtangg:
### What does this test do?

This tests is responsible for checking the integrity of the discard and deadletter mechanism implemented in the upkeep thread of taskbroker. An initial amount of messages is produced to kafka where half of the messages' on_attempts_exceeded is set to discard and the other half to deadletter. These messages have an `expires` value of 1 second. An extra message is produced to the topic which is to be completed by a taskworker first. This allows all the previous messages to be discarded/deadlettered by upkeep. During an interval, the upkeep thread collect all tasks that have expired, sets them all to a failed status, then appropriately discards or deadletters these messages. This process continues until all tasks have have a completed status (this means all tasks have either been discarded or deadlettered).

The test starts a taskworker and a taskbroker in separate threads. Synchronization events are use to instruct the taskworker when complete the last message and shutdown. Once the upkeep thread has successfully completed all messages in sqlite, taskbroker shuts down. Finally, this total number of completed messages and messages in the DLQ topic is validated.

### DIFF
--- a/python/integration_tests/helpers.py
+++ b/python/integration_tests/helpers.py
@@ -3,7 +3,7 @@ import subprocess
 import sqlite3
 import time
 
-from confluent_kafka import Producer
+from confluent_kafka import Consumer, Producer, KafkaException
 from pathlib import Path
 from uuid import uuid4
 from sentry_protos.taskbroker.v1.taskbroker_pb2 import (
@@ -29,6 +29,7 @@ class TaskbrokerConfig:
         db_path: str,
         max_pending_count: int,
         kafka_topic: str,
+        kafka_deadletter_topic: str,
         kafka_consumer_group: str,
         kafka_auto_offset_reset: str,
         grpc_port: int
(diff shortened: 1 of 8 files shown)


python/integration_tests/helpers.py
```diff
@@ -90,29 +102,64 @@ def serialize_task_activation(i, args: list, kwargs: dict) -> bytes:
...
+    """
+    try:
+        producer = Producer(TEST_PRODUCER_CONFIG)
+
+        for message in custom_messages:
+            task_message = message.SerializeToString()
+            producer.produce(topic_name, task_message)
+
+        producer.poll(5)  # trigger delivery reports
+        producer.flush()
```
evanh, line 135: I don't think you need the flush here, my understanding is that `flush` is a wrapper for `poll(0)`.

evanh (approved):

markstory (approved):

PR #147, opened 2025-01-30
feat(telemetry): A single place to configure log filtering

john-z-yang:
I was trying to figure out how to get rid of the annoying h2 logs
```
2025-01-30T07:17:43.333156Z DEBUG Connection: h2::codec::framed_read: received frame=Headers { stream_id: StreamId(7), flags: (0x4: END_HEADERS) } peer=Server
2025-01-30T07:17:43.333259Z DEBUG Connection: h2::codec::framed_read: received frame=WindowUpdate { stream_id: StreamId(7), size_increment: 5 } peer=Server
2025-01-30T07:17:43.333311Z DEBUG Connection: h2::codec::framed_read: received frame=Data { stream_id: StreamId(7), flags: (0x1: END_STREAM) } peer=Server
2025-01-30T07:17:43.334691Z DEBUG Connection: h2::codec::framed_write: send frame=Headers { stream_id: StreamId(7), flags: (0x5: END_HEADERS | END_STREAM) } peer=Server
```

After figuring it out I realize I'll need to create yet another thing in the config struct / file / envar to adjust the h2 logs, similar to what we do with rdkafka. This can quickly get confusing when trying to find the right filter for different parts of our codebase.

I learned that we can just control everything with a single config field / envar like
```
log_filter: debug,librdkafka=warn,h2=off
```
Which sets our log level to debug (or above), set the rdkafka log level to warning (or above) and turns off logs from h2.
This feels a lot more readable and maintainable to me. This also has the benefit of not needing to muck around with settings when we adopt another really verbose library in the future.

What do we think?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,15 +1,12 @@
 use figment::{
-    providers::{Env, Format, Serialized, Yaml},
+    providers::{Env, Format, Yaml},
     Figment, Metadata, Profile, Provider,
 };
 use rdkafka::ClientConfig;
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
-use crate::{
-    logging::{LogFormat, LogLevel},
-    Args,
-};
+use crate::{logging::LogFormat, Args};
 
 #[derive(PartialEq, Debug, Deserialize, Serialize)]
 pub struct Config {
@@ -22,11 +19,8 @@ pub struct Config {
     /// The sampling rate for tracing data.
     pub traces_sample_rate: Option<f32>,
 
-    /// The log level to filter application logging to.
-    pub log_level: LogLevel,
-
-    /// The log level to filter rdkafka logging to.
-    pub rdkafka_log_level: LogLevel,
+    /// The log filter to apply application logging to.
+    pub log_filter: String,
 
     /// The log format to use
     pub log_format: LogFormat,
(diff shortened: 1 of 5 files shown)


src/config.rs
```diff
@@ -22,11 +19,8 @@ pub struct Config {
     /// The sampling rate for tracing data.
     pub traces_sample_rate: Option<f32>,
 
-    /// The log level to filter application logging to.
-    pub log_level: LogLevel,
-
-    /// The log level to filter rdkafka logging to.
-    pub rdkafka_log_level: LogLevel,
+    /// The log filter to apply application logging to.
+    pub log_filter: String,
```
evanh, line 23: Could you maybe provide a link to docs on what this string should be?

src/logging.rs
```diff
@@ -119,7 +81,7 @@ pub fn init(log_config: LoggingConfig) {
     };
 
     let logs_subscriber = tracing_subscriber::registry()
-        .with(formatter.with_filter(log_config.log_level.level_filter()))
+        .with(formatter.with_filter(EnvFilter::new(log_config.log_filter)))
```
evanh, line 84: If this string is invalid, does this fail with a useful error message?

markstory (approved):
Having a general filter string should be a better solution than what we're doing now.

src/logging.rs
john-z-yang: Yes, it will parse the parseable ones and tell continue on.
```console
╰─❯ TASKBROKER_LOG_FILTER="debug,  BADD CAFE, 1BADB 002" cargo run
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/taskbroker`
ignoring `  BADD CAFE`: invalid filter directive
ignoring ` 1BADB 002`: invalid filter directive
2025-01-30T19:47:54.022263Z  INFO taskbroker: GRPC server listening on 0.0.0.0:50051
```
If nothing is parsable, it will default to `Error` level filtering.

src/config.rs
john-z-yang: Yup, updated!

evanh (approved):

PR #149, opened 2025-01-30
feat(telemetry): Add config to define metric default tag/values

john-z-yang:
Adds a `default_metric_tags` to our config so all metrics emitted will be tagged with those key/value pairs.
This is useful for emitting metrics with tags that distinguishes different pods since we're not running out datadog agent as a daemonset.

Use it like so

config file:
```yaml
default_metrics_tags:
    key_1: value_1
    key_2: value_2
```

envars:
```console
TASKBROKER_DEFAULT_METRICS_TAGS="{key_1=value_1,key_2=value_2}"
```

k8s (according to docs [here](https://kubernetes.io/docs/tasks/inject-data-application/define-environment-variable-container/#using-environment-variables-inside-of-your-config)):
```yaml
spec:
  containers:
  - name: taskbroker
    env:
    - name: TASKBROKER_DEFAULT_METRICS_TAGS
      value: "{host=$(HOSTNAME)}"
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -4,7 +4,7 @@ use figment::{
 };
 use rdkafka::ClientConfig;
 use serde::{Deserialize, Serialize};
-use std::borrow::Cow;
+use std::{borrow::Cow, collections::BTreeMap};
 
 use crate::{logging::LogFormat, Args};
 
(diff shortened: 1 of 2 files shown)


evanh (approved):
Did you test that this actually works? Also how would we actually use this to populate the hostname/podname?

john-z-yang:
> Did you test that this actually works? Also how would we actually use this to populate the hostname/podname?

I tested that the env vars and config files are read correctly, but we'll need to put it onto the sandbox to test if it does what it says on the tin. At this point I don't trust that metrics crate much.

PR #153, opened 2025-02-03
ref(grpc): Refactor set_task_status

john-z-yang:
I was debugging the `set_task_status` grpc code and I found it a little bit difficult to follow with the mutation and high level of nesting.

Let me know if this is actually easier to read, I am obviously biased


markstory (approved):
Looks good to me. The destructuring for `FetchNextTask` is much cleaner now 🧹

PR #151, opened 2025-01-31
chore(devservices): Bump devservices to 1.0.14

hubertdeng123:
This bumps devservices to version 1.0.14 since the version here seems to be pretty outdated.

PR #154, opened 2025-02-03
fix(logging): Add some logging to in between upkeep in DLQ integration test

enochtangg:
This integration test is flaky in CI. Since it's difficult to reproduce, this PR adds some logging to see what the current state of sqlite is at each time interval to see if upkeep is doing its job.


markstory (approved):

enochtangg:
This PR also reverts: https://github.com/getsentry/taskbroker/pull/151 due to backwards compatibility issues with resetting kafka which is causing this flaky test.

PR #155, opened 2025-02-03
fix(sqlite): Change id column from UUID to TEXT

evanh:
The UUID type is a generic type in SQLite, which automatically tries to convert values into numbers where possible. Since some UUIDs are also valid numbers, SQLite was converting those values to numbers. This was generally fine except for UUIDs that had just numbers, with a lower case 'e', and then more numbers, e.g. 97033619637148748591279e7866347. SQlite would interpret the `e` as an exponent character, and end up storing +Inf.

This is all avoided by storing the key as a TEXT string, which prevents this type of interpolation.

### DIFF
--- a/migrations/0001_create_inflight_taskactivations.sql
+++ b/migrations/0001_create_inflight_taskactivations.sql
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS inflight_taskactivations (
-    id UUID NOT NULL PRIMARY KEY,
+    id TEXT NOT NULL PRIMARY KEY,
     activation BLOB NOT NULL,
     partition INTEGER NOT NULL,
     offset BIGINTEGER NOT NULL,


markstory (approved):

enochtangg (approved):
lol, noice

migrations/0001_create_inflight_taskactivations.sql
```diff
@@ -1,5 +1,5 @@
 CREATE TABLE IF NOT EXISTS inflight_taskactivations (
-    id UUID NOT NULL PRIMARY KEY,
+    id TEXT NOT NULL PRIMARY KEY,
```
fpacifici, line 2: Any reason not to encode the UUID as 16 bytes binary blob? That would make the primary index smaller and comparisons quicker

evanh: Yes, unfortunately BLOBs have the same problem as UUID: they are interpreted numerically. So the value `97033619637148748591279e7866347` would also be encoded to `Inf` in a BLOB.

PR #158, opened 2025-02-03
chore(makefile): Use docker commands to reset kafka

hubertdeng123:
We shouldn't be using `devservices purge` for this command as it resets all devservices containers, impacting development environments in sentry/snuba/relay if used.


john-z-yang (approved):

enochtangg:
Close in favour for: https://github.com/getsentry/taskbroker/pull/154. Devservices version bump was reverted for now.

PR #161, opened 2025-02-04
feat(consumer): Improve consumer write throughput

john-z-yang:
Adds a batching step in front of the db write step.

The current Sqlite write step works like so:
1. When a message come in from the previou step, it queries to see if size(db) + size(buffer) is smaller than max pending tasks, if it is, add the message to the buffer.
2. When size(db) + size(buffer) is equal to the max pending tasks, or every 4 seconds, flush the buffer to the db

This worked reasonably well but it requires us to make a query each time we see a message. By separating the batching step before the writer, we only need to query the DB every `max_pending_buffer_count` messages.


markstory (approved):
Makes sense to me.

PR #163, opened 2025-02-04
fix(dlq-test): bump consumer timeout to 10 min

enochtangg:
Two minutes for the consumer to completely DLQ/discarding 10,000 tasks might have been over ambitious. Let's bump it to 10 minutes and see if it fixes flaky test.

### DIFF
--- a/python/integration_tests/test_upkeep_dlq.py
+++ b/python/integration_tests/test_upkeep_dlq.py
@@ -243,7 +243,7 @@ def test_upkeep_dlq() -> None:
     num_messages = 10_000
     num_partitions = 4
     max_pending_count = 100_000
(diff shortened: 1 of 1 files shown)


markstory (approved):

python/integration_tests/test_upkeep_dlq.py
```diff
@@ -243,7 +243,7 @@ def test_upkeep_dlq() -> None:
     num_messages = 10_000
     num_partitions = 4
     max_pending_count = 100_000
-    taskbroker_timeout = 120  # the time in seconds to wait for taskbroker to process
+    taskbroker_timeout = 600  # the time in seconds to wait for taskbroker to process
```
markstory, line 246: If we consistently go over 5min, we should consider reducing the number of messages.

PR #165, opened 2025-02-04
ref(store): Set synchronous mode to normal

john-z-yang:
According to the sqlite doc: www.sqlite.org/pragma.html#pragma_synchronous
> When synchronous is NORMAL (1), the SQLite database engine will still sync at the most critical moments, but less often than in FULL mode. There is a very small (though non-zero) chance that a power failure at just the wrong time could corrupt the database in [journal_mode](https://www.sqlite.org/pragma.html#pragma_journal_mode)=DELETE on an older filesystem. [WAL mode](https://www.sqlite.org/wal.html) is safe from corruption with synchronous=NORMAL, and probably DELETE mode is safe too on modern filesystems.

We are already using WAL mode from the look of the sqlx query debug output
```
2025-02-04T23:00:31.838496Z DEBUG sqlx::query: summary="PRAGMA journal_mode = WAL; …" db.statement="\n\nPRAGMA journal_mode = WAL; PRAGMA foreign_keys = ON; \n" rows_affected=0 rows_returned=1 elapsed=1.142834ms elapsed_secs=0.001142834
```
But I added it just to be more explicit.

### DIFF
--- a/src/inflight_activation_store.rs
+++ b/src/inflight_activation_store.rs
@@ -6,7 +6,10 @@ use prost::Message;
 use sentry_protos::taskbroker::v1::{OnAttemptsExceeded, TaskActivation, TaskActivationStatus};
 use sqlx::{
     migrate::MigrateDatabase,
-    sqlite::{SqliteConnectOptions, SqlitePool, SqliteQueryResult, SqliteRow},
+    sqlite::{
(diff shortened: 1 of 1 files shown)


john-z-yang:
Really weird that this is causing a lower than expected amount in the sqlite library for the consumer rebalancing test

PR #166, opened 2025-02-05
perf(store): Add benchmarks

john-z-yang:
Adds a benchmark for simulating worker ⇔ store interactions.

Use `make bench` or `cargo bench` to run.

```
bench_InflightActivationStore/process_activations
                        time:   [13.091 s 13.313 s 13.525 s]
                        thrpt:  [739.39  elem/s 751.15  elem/s 763.88  elem/s]
                 change:
                        time:   [-21.574% -19.199% -16.927%] (p = 0.00 < 0.05)
                        thrpt:  [+20.376% +23.760% +27.508%]
                        Performance has improved.
```


benches/inflight_activation_store_bench.rs
```diff
@@ -0,0 +1,77 @@
...
+
+            while let Some(ref activation) = pending {
+                store
+                    .set_status(
+                        &activation.activation.id,
+                        InflightActivationStatus::Complete,
+                    )
+                    .await
+                    .unwrap();
+                sleep(Duration::from_millis(50)).await;
```
markstory, line 42: Why do you want to sleep in a benchmark test?

john-z-yang: So I tried letting it run without a sleep, that gives me this error when I have a large number of workers (>16)

```
'tokio-runtime-worker' panicked at benches/inflight_activation_store_bench.rs:32:18:
called `Result::unwrap()` on an `Err` value: error returned from database: (code: 5) database is locked

Caused by:
    (code: 5) database is locked
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
This seems like a very strange error to me as I expect if the db is contended, the .await call would just wait.

Some cursory googling suggests that this is because sqlite is actually single threaded under the hood for writes, and for those 2 calls we're doing writes with `UPDATE`.

The way I can get around with this error is to force only a single connection when we initialize the pool by pinning the max connection to 1 when the store is being initialized
```rust
let sqlite_pool = PoolOptions::new().max_connections(1).connect(url).await?;
```

Makefile
```diff
@@ -89,6 +89,13 @@ test-upkeep-expiry: build reset-kafka ## Run the upkeep expiry integration test
 integration-test: test-rebalance test-worker-processing test-upkeep-retry ## Run all integration tests
 .PHONY: integration-test
 
+
+# Benchmarks
+
+bench::
```
markstory, line 95:
```suggestion
bench:
```

I don't think `::` is valid Make.

markstory (commented):
Looking good so far.

enochtangg (approved):
Looks good, this seems like a great lightweight way to benchmark taskbroker and check if we introduced performance regressions. In the future, I'd be nice to also make benchmarks for upkeep and the grpc server.

Issue #170, opened 2025-02-06
Add expires_at to sqlite schema and revise logic for activations with expires


markstory:
As part of the work to address edge cases in full databases, we will be deprecating `remove_at` and replacing it with an processing attempt counter and `expires_at` time.

We'll need to:

- Add `expires_at` to the sqlite schema
- When an activation has `expires` set, we should calculate the `expires_at` time based on the activations `received_at + expires`. We want to use `received_at` because expires times are based on *spawn time* not *processing time*.
- Remove the logic that sets `remove_at` based on `expires`.

Upkeep logic and `get_pending_activations` will also need to be updated.

- Activations past `expires_at` should not be fetched with `GetTask`
- Activations past `expires_at` should be discarded during upkeep.

PR #173, opened 2025-02-07
feat(expires): Track task expiry explicitly

evanh:
Currently if a task has an expiry set, that expiry date is tracked through the remove_at column. However that column is also used to track tasks that aren't being processed because of worker failures.

The issue with that is that expired tasks should be handled differently from tasks that are not being processed. Expired tasks can be removed immediately as soon as the expiry has passed, regardless of any other tasks. Tasks that aren't being processed might be in that state because the workers aren't running, so they can't be removed until other tasks start being completed.

Add an explicit column for expires_at, that is only populated if the task has an expiry. Also update the upkeep thread to mark expired tasks for removal.

While in theory a check could be added to the consumer to avoid even ingesting tasks that are expired, letting the upkeep thread remove them means the tasks get properly discarded/retried/dlq.

Fixes https://github.com/getsentry/taskbroker/issues/170

### DIFF
--- a/src/consumer/deserialize_activation.rs
+++ b/src/consumer/deserialize_activation.rs
@@ -45,22 +45,21 @@ pub fn new(
         let now = Utc::now();
 
         // Determine the deadletter_at time using config and activation expires time.
-        let mut remove_at = now.add(config.remove_deadline);
+        let mut expires_at = None;
         if let Some(expires) = activation.expires {
             let expires_duration = Duration::from_secs(expires);
-            if expires_duration < config.remove_deadline {
-                // Expiry times are based on the time the task was received
-                // not the time it was dequeued from Kafka.
-                let activation_received = activation.received_at.map_or(now, |ts| {
-                    match Utc.timestamp_opt(ts.seconds, ts.nanos as u32) {
-                        MappedLocalTime::Single(ts) => ts,
-                        _ => now,
-                    }
-                });
-                remove_at = activation_received + expires_duration;
-            }
+            // Expiry times are based on the time the task was received
+            // not the time it was dequeued from Kafka.
+            let activation_received = activation.received_at.map_or(now, |ts| {
+                match Utc.timestamp_opt(ts.seconds, ts.nanos as u32) {
+                    MappedLocalTime::Single(ts) => ts,
+                    _ => now,
+                }
+            });
+            expires_at = Some(activation_received + expires_duration);
         }
 
+        let remove_at = now.add(config.remove_deadline);
         Ok(InflightActivation {
             activation,
             status: InflightActivationStatus::Pending,
@@ -69,6 +68,7 @@ pub fn new(
             added_at: Utc::now(),
             processing_deadline: None,
             remove_at,
+            expires_at,
             at_most_once,
             namespace,
         })
--- a/src/inflight_activation_store.rs
+++ b/src/inflight_activation_store.rs
@@ -73,6 +73,9 @@ pub struct InflightActivation {
     /// depending on the retry policy of an activation it will either be deadlettered or discarded.
     pub remove_at: DateTime<Utc>,
 
+    // If the task has specified an expiry, this is the timestamp after which the task should be removed from inflight store
+    pub expires_at: Option<DateTime<Utc>>,
+
     /// The timestamp for when processing should be complete
     pub processing_deadline: Option<DateTime<Utc>>,
 
@@ -109,6 +112,7 @@ struct TableRow {
     offset: i64,
     added_at: DateTime<Utc>,
     remove_at: DateTime<Utc>,
+    expires_at: Option<DateTime<Utc>>,
     processing_deadline_duration: u32,
     processing_deadline: Option<DateTime<Utc>>,
     status: InflightActivationStatus,
@@ -127,6 +131,7 @@ impl TryFrom<InflightActivation> for TableRow {
             offset: value.offset,
             added_at: value.added_at,
             remove_at: value.remove_at,
+            expires_at: value.expires_at,
             processing_deadline_duration: value.activation.processing_deadline_duration as u32,
             processing_deadline: value.processing_deadline,
             status: value.status,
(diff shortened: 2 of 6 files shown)


src/inflight_activation_store/store_tests.rs
```diff
@@ -0,0 +1,762 @@
+use std::collections::{HashMap, HashSet};
```
evanh, line 1: I'm going to open a separate PR that also moves the tests, so that this PR is cleaner.

src/consumer/deserialize_activation.rs
```diff
@@ -45,22 +45,21 @@ pub fn new(
...
-                    match Utc.timestamp_opt(ts.seconds, ts.nanos as u32) {
-                        MappedLocalTime::Single(ts) => ts,
-                        _ => now,
-                    }
-                });
-                remove_at = activation_received + expires_duration;
-            }
+            // Expiry times are based on the time the task was received
+            // not the time it was dequeued from Kafka.
+            let activation_received = activation.received_at.map_or(now, |ts| {
```
john-z-yang, line 53: I don't think we need this map_or here since the schema does not define [recieved_at](https://github.com/getsentry/sentry-protos/blob/main/proto/sentry_protos/taskbroker/v1/taskbroker.proto#L54) as optional

src/consumer/deserialize_activation.rs
```diff
@@ -45,22 +45,21 @@ pub fn new(
         let now = Utc::now();
 
         // Determine the deadletter_at time using config and activation expires time.
-        let mut remove_at = now.add(config.remove_deadline);
+        let mut expires_at = None;
```
john-z-yang, line 48: Nit, I think it's clearer if assign it directly instead of mutating this variable here.
i.e.
```rust
let mut expires_at = if let Some(expires) = activation.expires {
        ...
    } else {
        None
    } 
```

src/inflight_activation_store/store_tests.rs
john-z-yang: I think you might want to do
```rust
#[cfg(test)]
mod tests {
...
}
```
to surround the tests, otherwise they make it into the production binary

src/consumer/deserialize_activation.rs
markstory: Nested structs are always optionals in rust bindings :cry:

src/inflight_activation_store.rs
```diff
@@ -203,6 +209,7 @@ impl InflightActivationStore {
                 b.push_bind(row.offset);
                 b.push_bind(row.added_at);
                 b.push_bind(row.remove_at);
+                b.push_bind(row.expires_at);
```
markstory, line 212:
```suggestion
                b.push_bind(row.expires_at.timestamp());
```
without the `.timestamp()` sqlx will convert this into a datetime string, and jam it into the INTEGER column, which SQLite happily accepts.

src/inflight_activation_store.rs
```diff
@@ -237,6 +244,9 @@ impl InflightActivationStore {
             WHERE status = ",
         );
         query_builder.push_bind(InflightActivationStatus::Pending);
+        query_builder.push(" AND (expires_at IS NULL OR expires_at > ");
+        query_builder.push_bind(now);
```
markstory, line 248:
```suggestion
        query_builder.push_bind(now.timestamp());
```

src/inflight_activation_store.rs
```diff
@@ -416,6 +427,29 @@ impl InflightActivationStore {
...
+    pub async fn handle_expires_at(&self) -> Result<u64, Error> {
+        let now = Utc::now();
+        let update_result = sqlx::query(
+            r#"UPDATE inflight_taskactivations
+            SET status = $1
+            WHERE expires_at < $2 AND status = $3
+            "#,
+        )
+        .bind(InflightActivationStatus::Failure)
+        .bind(now)
```
markstory, line 445:
```suggestion
        .bind(now.timestamp())
```

src/consumer/deserialize_activation.rs
evanh: It's an Option in the proto rust file. We have an older version though. I think I'll save this change for a separate PR.

john-z-yang: Yeah, do you think we should `.unwrap` or `.expect` here? I'm not strongly opinionated on this, but I feel like maybe we can `.unwrap` or `.expect`on fake optionals that is created by the bindings and use the monads (map_or/unwrap_or/etc) in real optionals?

src/inflight_activation_store.rs
evanh: Why is this just a problem for this column? Why don't we have that issue for `remove_at` and `expire_at`?

markstory: We do have issues with those columns as well. They currently get isoformat strings stored despite having an integer storage type.

src/consumer/deserialize_activation.rs
markstory: I think using `.unwrap()` on binding optionals is reasonable. While they could theoretically be empty they should never be in-practice.

markstory (approved):
Looks good once the `timestamp()` calls are applied.

src/inflight_activation_store.rs
```diff
@@ -73,6 +73,9 @@ pub struct InflightActivation {
     /// depending on the retry policy of an activation it will either be deadlettered or discarded.
     pub remove_at: DateTime<Utc>,
 
+    // If the task has specified an expiry, this is the timestamp after which the task should be removed from inflight store
```
markstory, line 76:
```suggestion
    /// If the task has specified an expiry, this is the timestamp after which the task should be removed from inflight store
```

The extra / makes it a docstring that shows up in LSP.

markstory (approved):

src/inflight_activation_store.rs
```diff
@@ -189,8 +195,9 @@ impl InflightActivationStore {
         }
         let mut query_builder = QueryBuilder::<Sqlite>::new(
             "INSERT INTO inflight_taskactivations \
-            (id, activation, partition, offset, added_at, remove_at, processing_deadline_duration, processing_deadline, status, at_most_once, namespace)",
+            (id, activation, partition, offset, added_at, remove_at, expires_at, processing_deadline_duration, processing_deadline, status, at_most_once, namespace)",
         );
+        println!("batch: {:?}", batch);
```
enochtangg, line 200: 💣

enochtangg (approved):

PR #176, opened 2025-02-07
fix: Fix timestamp math comparing datetime strings + timestamps

markstory:
When I converted the DATETIME columns to INTEGER types, I missed some `datetime()` calls. I also incorrectly assumed that sqlx would convert `Utc` into timestamp values - it does not.

I had to make a bunch of adjustments to tests as they were working by happy accident. These changes will conflict with the other work we have in flight to remove timestamp column usage, so I'll rebase and merge once those changes have landed.


john-z-yang:
I am not familiar with this part of our codebase, but would it be less error prone if we change the `TableRow` struct from
```
#[derive(Debug, FromRow)]
struct TableRow {
    ...
    added_at: DateTime<Utc>,
    remove_at: DateTime<Utc>,
    ...
}
```
to
```
#[derive(Debug, FromRow)]
struct TableRow {
    ...
    added_at: u64
    remove_at: u64,
    ...
}
```
And we make the conversion from DateTime<Utc> → u64 only via
```
impl TryFrom<InflightActivation> for TableRow {
```

src/upkeep.rs
```diff
@@ -281,6 +281,7 @@ mod tests {
             on_attempts_exceeded: OnAttemptsExceeded::Discard as i32,
             at_most_once: None,
         });
+        records[1].added_at = records[1].added_at.add(Duration::from_secs(1));
```
evanh, line 284:
```suggestion
        records[1].added_at = records[1].added_at + Duration::from_secs(1);
```

Then you don't need to import the `std::ops::Add`.

evanh (approved):

john-z-yang:
> > I am not familiar with this part of our codebase, but would it be less error prone
> 
> Yeah that would help. I can make that change as a follow up.

I can make the PR if you like, I want to get more familiarity with this part of the code!

markstory:
@john-z-yang That would be great!

PR #177, opened 2025-02-10
fix: Use STRICT when building SQLite table

evanh:
This makes SQLite enforce that the types of the columns are restricted just to the ones that are
explicitly supported, and does validation on the values being inserted to ensure they are the
correct type.


markstory (approved):
Nice :clap: This should help keep our timestamp logic more reliable and well behaved.

PR #179, opened 2025-02-10
feat(processing-attempts): Replace remove_at with processing_attempts

enochtangg:
There is an edge case with the `remove_at` mechanism which can cause taskbroker to halt. This column is not enough for the taskbroker to differentiate activations that are stuck in sqlite because workers are unable to process them, and activations that are stuck due to workers not being available. 

This PR is responsible for replacing the `remove_at` column with `processing_attempts` which counts the number of times an activation has been reset back from `processing` to `pending` due to processing_deadline exceeding. When `processing_attempts` >= `max_processing_attempts` (defined in configuration), the new `handle_processing_attempts` function in upkeep will move the task to `failed`. Finally, the `handle_failed_tasks` cleans them up.

In this implementation, the `processing_attempts` is reset for a task is retried (reproduced into the topic). This means there is a scenario where a task with a `max_processing_attempts=3` and `retry=5` could be attempted 15 times assuming the worker dies every time.

### DIFF
--- a/src/inflight_activation_store.rs
+++ b/src/inflight_activation_store.rs
@@ -242,8 +242,6 @@ impl InflightActivationStore {
         &self,
         namespace: Option<&str>,
     ) -> Result<Option<InflightActivation>, Error> {
-        let now = Utc::now();
-
         let mut query_builder = QueryBuilder::new(
             "UPDATE inflight_taskactivations
             SET processing_deadline = unixepoch('now', '+' || processing_deadline_duration || ' seconds'), status = ",
@@ -500,40 +498,17 @@ impl InflightActivationStore {
         Ok(result.rows_affected())
     }
 
-    /// Remove completed tasks that do not have an incomplete record following
-    /// them in offset order.
+    /// Remove completed tasks
     ///
     /// This method is a garbage collector for the inflight task store.
     pub async fn remove_completed(&self) -> Result<u64, Error> {
         let mut atomic = self.sqlite_pool.begin().await?;
 
-        let incomplete_query = sqlx::query(
-            r#"
-            SELECT "added_at"
-            FROM inflight_taskactivations
-            WHERE status != $1
-            ORDER BY "added_at"
-            LIMIT 1
-            "#,
-        )
-        .bind(InflightActivationStatus::Complete)
-        .fetch_optional(&mut *atomic)
-        .await?;
-
-        let earliest_incomplete_added_at: DateTime<Utc> =
-            if let Some(query_result) = incomplete_query {
-                query_result.get("added_at")
-            } else {
-                return Ok(0);
-            };
-
-        let cleanup_query = sqlx::query(
-            r#"DELETE FROM inflight_taskactivations WHERE status = $1 AND "added_at" < $2"#,
-        )
-        .bind(InflightActivationStatus::Complete)
-        .bind(earliest_incomplete_added_at.timestamp())
-        .execute(&mut *atomic)
-        .await?;
+        let cleanup_query =
+            sqlx::query(r#"DELETE FROM inflight_taskactivations WHERE status = $1"#)
+                .bind(InflightActivationStatus::Complete)
+                .execute(&mut *atomic)
+                .await?;
 
         atomic.commit().await?;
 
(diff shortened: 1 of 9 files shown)


enochtangg:
Depends on: https://github.com/getsentry/taskbroker/pull/173. There's a test I had to commented out for now. After that PR is merged, I will rebase this branch and update the test.

enochtangg:
Rebased main with `expires_at` changes. Ready for review.

src/inflight_activation_store.rs
```diff
@@ -244,10 +263,10 @@ impl InflightActivationStore {
             WHERE status = ",
         );
         query_builder.push_bind(InflightActivationStatus::Pending);
-        query_builder.push(" AND (expires_at IS NULL OR expires_at > ");
-        query_builder.push_bind(now.timestamp());
+        query_builder.push(" AND (processing_attempts < ");
```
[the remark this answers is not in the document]
enochtangg: Good idea 👍 Updated `handle_processing_deadline` to fail tasks that have exceeded the max processing attempts in the same transaction.

src/inflight_activation_store.rs
```diff
@@ -384,47 +403,29 @@ impl InflightActivationStore {
...
+        let result = sqlx::query(
+            "UPDATE inflight_taskactivations SET status = $1 WHERE status = $2 AND processing_attempts >= $3",
         )
         .bind(InflightActivationStatus::Failure)
-        .bind(now.timestamp())
-        .bind(max_added_at.timestamp())
         .bind(InflightActivationStatus::Pending)
-        .execute(&mut *atomic)
-        .await?;
+        .bind(self.config.max_processing_attempts as i32)
```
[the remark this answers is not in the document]
enochtangg: I kinda like how all the InflightActivationStore methods can be called independently without any context params. It makes reading the upkeep code a bit cleaner. We have same pattern we have for `InflightActivationWriter`. Though, I'm not super opinionated on this. Let me know what you think!

src/inflight_activation_store/store_tests.rs
```diff
@@ -111,7 +132,12 @@ async fn test_store_duplicate_id_between_batches() {
 #[tokio::test]
 async fn test_get_pending_activation() {
     let url = generate_temp_filename();
-    let store = InflightActivationStore::new(&url).await.unwrap();
+    let store = InflightActivationStore::new(
+        &url,
+        InflightActivationStoreConfig::from_config(&create_integration_config()),
+    )
+    .await
+    .unwrap();
```
markstory, line 140: Should we have a test_utils method that builds store instances now that the setup is a bit more involved?

evanh (approved):
This looks good, nicely done!

src/upkeep.rs
```diff
@@ -125,32 +125,34 @@ pub async fn do_upkeep(
...
+        result_context.processing_deadline_reset = counts.0;
+        result_context.processing_attempts_exceeded = counts.1;
     }
 
-    if let Ok(expired_count) = store
-        .handle_expires_at()
-        .instrument(info_span!("handle_expires_at"))
+    // 5.Handle tasks whos processing_attempts have exceeded max_processing_attempts
+    if let Ok(processing_attempts_exceeded) = store
+        .handle_processing_attempts()
```
markstory, line 139: Now that you're advancing activations to failure inside `handle_processing_deadline` wouldn't this be a no-op?

src/inflight_activation_store/store_tests.rs
enochtangg: Added util function for this 👍

src/upkeep.rs
enochtangg: That's true, removed `handle_processing_attempts` in favour for `handle_processing_deadline`.

src/inflight_activation_store.rs
```diff
@@ -359,72 +379,74 @@ impl InflightActivationStore {
...
+        let mut modified_rows = 0;
 
-        let now = Utc::now();
-        let update_result = sqlx::query(
-            r#"UPDATE inflight_taskactivations
-            SET status = $1
-            WHERE remove_at <= $2 AND added_at <= $3 AND status = $4
-            "#,
+        let result = sqlx::query(
+            "UPDATE inflight_taskactivations SET status = $1 WHERE status = $2 AND processing_attempts >= $3",
```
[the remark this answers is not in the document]
enochtangg: Removed function, see below

markstory (approved):
Nice work :clap:

PR #180, opened 2025-02-11
feat(kafka): update reducer interface

john-z-yang:
Reducer flush now returns
```rust
Result<Option<Self::Output>, ...>
```
instead of 
```rust
Result<Self::Output, ...>
```
So that it can signal that it yielded no value after a flush call.

This is useful for things like "the reducer is full but the db is also full", so the flush is a noop, the offsets do not get committed.


evanh (approved):

evanh (approved):
LGTM

PR #181, opened 2025-02-11
fix(test): fix flaky test

john-z-yang:
I think we're rounding the wrong value when comparing the 2 timestamps, hence why things are off by one sometimes.


john-z-yang:
Feel free to just merge this whenever you approve it. This test is getting on my nerves.

evanh (approved):

PR #185, opened 2025-02-11
chore(devservices): Bump devservices to 1.0.16

hubertdeng123:
Picks up quite a few changes, this bumps devservices to latest:
https://github.com/getsentry/devservices/releases/tag/1.0.16


john-z-yang:
I can't remember what we decided on, but I think there's actually a breaking change here so the makefile need to be updated, right?

hubertdeng123:
> I can't remember what we decided on, but I think there's actually a breaking change here so the makefile need to be updated, right?

Yep, thought it was added already from before but guess not

markstory (approved):

enochtangg (approved):

PR #186, opened 2025-02-12
fix(config) Reduce the default upkeep interval

markstory:
With most of our time math only having second resolution there isn't much reason to run upkeep more frequently than once a second, as the only step that will apply is the completed task garbage collection.


markstory:
> Would this actually be "increasing" the interval instead of reducing? Sort of ambiguous.

Yeah, increase the interval, decrease the frequency upkeep runs 🫠

PR #187, opened 2025-02-12
feat(expires_at): Refactor test upkeep expiry integration test

enochtangg:
Now that we have separated expires_at and processing_attempts (previously known as remove_at), this test needs to be updated to reflect those changes. This PR is responsible for the following:
* Rename `test_upkeep_dlq` to `test_upkeep_expiry`. The test doesn't only validate DLQing, it validates handle_expires_at which later triggers discarding and DLQing.
* Remove the need for last buffer message and taskworker thread in the integration test. This is because we no longer require an earliest completed task in order to expire the tasks before it. 
* Add some better logging to debug test when it fails.,


python/integration_tests/test_upkeep_expiry.py
```diff
@@ -118,26 +71,29 @@ def manage_taskbroker(
...
                     f"[taskbroker_0]: Finishing writting all {num_messages} "
-                    "task(s) to sqlite. Sending signal to taskworker to "
-                    "start processing"
+                    "task(s) to sqlite."
                 )
-                tasks_written_event.set()
+                finished_writing_tasks = True
             time.sleep(1)
 
+        print(finished_writing_tasks)
```
evanh, line 86: ?

evanh (approved):

markstory (approved):

PR #188, opened 2025-02-12
fix(tests): Refactor grpc tests

evanh:
Make these proper unit tests, which allows using the test_utils functions and provide actual
 coverage of the underlying functions.

This change moves the file into the grpc module, as well as add tests that successfully fetch,
update and refetch tasks.

Also rename the integration tests file for consistency.


evanh:
Depends on https://github.com/getsentry/taskbroker/pull/184

PR #190, opened 2025-02-14
feat(store): Add option to configure sqlite sync mode

john-z-yang:


src/config.rs
```diff
@@ -115,6 +118,7 @@ impl Default for Config {
             kafka_auto_offset_reset: "latest".to_owned(),
             kafka_send_timeout_ms: 500,
             db_path: "./taskbroker-inflight.sqlite".to_owned(),
+            db_sync_mode: "normal".to_owned(),
```
john-z-yang, line 121: Note that this defaults to normal, which is different from full. So once this is merged, the default will become normal

enochtangg (approved):

PR #191, opened 2025-02-14
fix: flush WAL on each commit to the store

markstory:
With sync=normal we *could* lose entries in the WAL if the machine were to die abruptly. By forcing a sync on each batch write we can ensure that we won't lose tasks between kafka and sqlite.


evanh (approved):

evanh (commented):
Could we make this configurable?

enochtangg (approved):

PR #193, opened 2025-02-18
fix(upkeep): Tweak upkeep producer settings

evanh:
The upkeep currently takes much much longer to run when it has to retry/deadletter messages into the
local producer. In an effort to improve that, have the local producer flush messages based on time
instead of on batch size.


markstory (approved):

evanh:
This isn't working well and there is a better option for this (batching futures).

PR #195, opened 2025-02-19
ref(upkeep): Send retry in batches

john-z-yang:


src/upkeep.rs
```diff
@@ -100,23 +100,40 @@ pub async fn do_upkeep(
...
+                            Timeout::After(Duration::from_millis(config.kafka_send_timeout_ms)),
+                        )
+                        .await;
+                    match send_result {
+                        Ok(_) => Ok(inflight.activation.id),
+                        Err(err) => Err(err),
+                    }
+                }
+            })
+            .collect::<JoinSet<_>>();
```
john-z-yang, line 123: I wonder if this really should be a JoinSet, maybe it should just be a vec of future instead

src/upkeep.rs
```diff
@@ -100,22 +100,38 @@ pub async fn do_upkeep(
...
+                    match send_result {
+                        Ok(_) => Ok(inflight.activation.id),
+                        Err(err) => Err(err),
+                    }
+                }
+            })
+            .collect::<Vec<_>>();
+
+        let mut ids = Vec::new();
+        for future in futures {
```
evanh, line 126: Couldn't this do `join_all`?

john-z-yang: We discussed offline that this is functionally the same as using `FuturesUnordered`

john-z-yang:
@evanh Feel free to just merge this if you approve it in the morning if you want to run experiment with this change

evanh (approved):

PR #200, opened 2025-02-21
feat: Add authentication middleware

markstory:
Implement basic request signature verification with support for key rotation. I still need to add tests for the `validate_signature` method.


src/auth_middleware.rs
```diff
@@ -0,0 +1,132 @@
...
+use crate::config::Config;
+
+type Hmac256 = Hmac<Sha256>;
+
+#[derive(Debug, Clone, Default)]
+pub struct AuthLayer {
+    pub shared_secret: Vec<String>,
+}
+
+// TODO do I need the layer?
```
evanh, line 23: Do you?

src/auth_middleware.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    fn call(&mut self, req: http::Request<TonicBody>) -> Self::Future {
+        let secret = self.secret.clone();
+        let mut inner = self.inner.clone();
+        mem::swap(&mut inner, &mut self.inner);
+
+        Box::pin(async move {
+            let (parts, body) = req.into_parts();
+            let body_bytes = body.collect().await.unwrap().to_bytes();
```
evanh, line 77: So easy!

markstory: Yeah, getting the futures to be compatible and types to line up was the tricky part. 😵

src/auth_middleware.rs
markstory: Yes, we need a Layer as the AuthService can't be connected to the server directly as it doesn't have any routes.

PR #201, opened 2025-02-24
chore(settings): Remove unused max-processing-deadline config

enochtangg:
The `max_processing_deadline` default is not used anywhere in taskbroker. Instead the default for this value is defined when the activation is first created (in sentry) [here](https://github.com/getsentry/sentry/blob/4c1f5d4bcf580e6f5727e25e23cce216d9f98fa7/src/sentry/taskworker/constants.py#L1).

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -79,10 +79,6 @@ pub struct Config {
     /// before being written to InflightTaskStore (sqlite).
     pub max_pending_buffer_count: usize,
 
-    /// The maximum value for processing deadlines. If a task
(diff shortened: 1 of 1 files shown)


evanh (commented):
If a task didn't have this set in Sentry, what would happen to that task in the taskbroker? Could it sit in the queue forever?

enochtangg:
@evanh If a task does not explicitly set this value, it would get defaulted to whatever was set on the namespace level. If nothing is set there, then it will get defaulted to the value I linked above.

evanh (approved):

PR #206, opened 2025-02-25
feat(upkeep): Separate processing deadline attempts

enochtangg:
When upkeep needs to handle a large number of deadlines and discards at once, this locks up sqlite for quite a bit. As a result, this is causing a significant decrease in throughput in our sandbox testing. This PR splits out the processing attempts work out of the processing deadline call to sqlite in hopes to reduce the total time sqlite is blocked for when we see a large influx of tasks that need to be discarded after a restart.

### DIFF
--- a/python/integration_tests/test_upkeep_expiry.py
+++ b/python/integration_tests/test_upkeep_expiry.py
@@ -231,11 +231,11 @@ Running test with the following configuration:
         for i in range(num_messages):
             if i % 2 == 0:
                 task_activation = generate_task_activation(
-                    OnAttemptsExceeded.ON_ATTEMPTS_EXCEEDED_DISCARD, 3
+                    OnAttemptsExceeded.ON_ATTEMPTS_EXCEEDED_DISCARD, 15
                 )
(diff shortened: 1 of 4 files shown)


evanh (approved):

python/integration_tests/test_upkeep_expiry.py
```diff
@@ -231,11 +231,11 @@ def test_upkeep_expiry() -> None:
         for i in range(num_messages):
             if i % 2 == 0:
                 task_activation = generate_task_activation(
-                    OnAttemptsExceeded.ON_ATTEMPTS_EXCEEDED_DISCARD, 3
+                    OnAttemptsExceeded.ON_ATTEMPTS_EXCEEDED_DISCARD, 15
```
evanh, line 234: Why did you change these?

enochtangg: The integration test was flakey. Tasks were expiring before all tasks were even inserted into sqlite which was failing this test.

PR #216, opened 2025-02-28
feat: Add incremental vacuum

markstory:
Doing many writes/deletes to sqlite can result in a fragmented database that may be underperforming. By running sqlite's incremental_vacuum we can reclain the empty pages and compact the database and indexes.

### DIFF
--- a/src/store/inflight_activation.rs
+++ b/src/store/inflight_activation.rs
@@ -1,4 +1,4 @@
-use std::str::FromStr;
+use std::{str::FromStr, time::Instant};
 
 use anyhow::{Error, anyhow};
 use chrono::{DateTime, Utc};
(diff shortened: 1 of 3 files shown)


enochtangg (commented):
I think we'd also need to enable `PRAGMA auto_vacuum=incremental` mode for this to actually do things right? Maybe we can make that configurable

evanh (approved):
Could you track the duration of this too? Would be good to know.

src/store/inflight_activation.rs
```diff
@@ -189,6 +189,7 @@ impl InflightActivationStore {
         let conn_options = SqliteConnectOptions::from_str(url)?
             .journal_mode(SqliteJournalMode::Wal)
             .synchronous(SqliteSynchronous::Normal)
+            .auto_vacuum(SqliteAutoVacuum::Incremental)
```
markstory, line 192: @enochtangg This should be emitting the pragma to enable incremental vacuum.

PR #222, opened 2025-03-03
feat: Add test coverage collection

markstory:
- Add llvm-cov and a new github action to collect coverage data


enochtangg (approved):
Looks good. This should fix the CI test: https://github.com/getsentry/taskbroker/pull/224. We should add the auth token to the coverage job too.

PR #227, opened 2025-03-05
feat(metrics): add latency metric between sentry and worker

john-z-yang:


src/grpc/server.rs
```diff
@@ -33,15 +33,29 @@ impl ConsumerService for TaskbrokerServer {
 
         match inflight {
             Ok(Some(inflight)) => {
+                if let Some(sentry_ts) = DateTime::from_timestamp(
+                    inflight.activation.received_at.unwrap().seconds,
+                    inflight.activation.received_at.unwrap().nanos as u32,
+                ) {
+                    metrics::histogram!("grpc_server.sentry_to_disptach.latency").record(
```
markstory, line 40:
```suggestion
                    metrics::histogram!("grpc_server.received_to_gettask.latency").record(
```

:shrug:

PR #236, opened 2025-03-07
pref(store): Separate sqlite into read and write pool

john-z-yang:
Sqlite allows multiple readers but a single writer, since we are a write heavy application we set our connection pool max connection to 1 to prevent `database is locked` error. But a small percentage of our query is actually only read, and with WAL we can do them concurrently. This pr separates the pool into a read and write pool so these queries can be concurrent.


markstory (approved):
Looks good to me.

Issue #232, opened 2025-03-07
Docker images lack version number


markstory:
When we build docker images for taskbroker, the `VERSION` file is empty and no data can be read by `get_version()` during runtime.

markstory:
The problem was related to ghcr and cloudbuild having different build scripts and the ghcr script missing a BUILD_ARG.

PR #246, opened 2025-03-12
fix: Rename variables used to build images

markstory:
Rename these values as they were using the old project name.


enochtangg (approved):

markstory:
Refs #232

PR #251, opened 2025-03-13
feat(settings): Add periodically updating runtime config options

enochtangg:
This PR is responsible for adding runtime configs to taskbroker. The path to the yaml file is defined under `Config.runtime_config_path`. This can be used to override with a ConfigMap in k8s. Otherwise, defaults are defined under `runtime_config_defaults.yaml`. The runtime config manager reloads the configuration file into memory as part of the maintenance task (which currently runs every 60 seconds).


src/main.rs
```diff
@@ -99,6 +103,7 @@ async fn main() -> Result<(), Error> {
                     _ = timer.tick() => {
                         let _ = maintenance_store.vacuum_db().await;
                         info!("ran maintenance vacuum");
+                        let _ = runtime_config.reload_config().await;
```
markstory, line 106: Could we update the log message to indicate that config is reloaded as well? It would be good if we could return a `Result` here that indicates if the config has changed. If we had that we could clean out killed tasks only when configuration is changed.

src/runtime_config.rs
```diff
@@ -0,0 +1,79 @@
...
+            config: RwLock::new(runtime_config),
+            path,
+        }
+    }
+
+    fn read_yaml_file(path: &str) -> RuntimeConfig {
+        let contents = std::fs::read_to_string(path)
+            .unwrap_or_else(|_| panic!("Failed to read config file from {}", path));
+        serde_yaml::from_str(&contents)
+            .unwrap_or_else(|_| panic!("Failed to parse YAML from {}", path))
```
markstory, line 29: We shouldn't panic here. If we push a bad configuration, I would much prefer we emit an error to sentry or logging and ignore the invalid configuration. If we panic, a bad config change will terminate our brokers.

src/main.rs
```diff
@@ -50,6 +51,9 @@ async fn log_task_completion(name: &str, task: JoinHandle<Result<(), Error>>) {
 async fn main() -> Result<(), Error> {
     let args = Args::parse();
     let config = Arc::new(Config::from_args(&args)?);
+    let runtime_config = Arc::new(RuntimeConfigManager::new(
+        config.runtime_config_path.clone(),
+    ));
```
markstory, line 56: Will this get pushed into the `InflightActivationStore`?

src/runtime_config.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
```
john-z-yang, line 30: I feel like this function should use the asynchronous counterpart from [tokio io](https://tokio.rs/tokio/tutorial/io). Right now it is fully synchronous which means this will take up a POSIX thread until the IO is completed.

src/runtime_config.rs
```diff
@@ -0,0 +1,79 @@
...
+#[derive(Debug, Deserialize, Clone)]
+pub struct RuntimeConfig {
+    /// A list of tasks to drop before inserting into sqlite.
+    pub drop_task_killswitch: Vec<String>,
+}
+
+pub struct RuntimeConfigManager {
+    pub config: RwLock<RuntimeConfig>,
+    pub path: String,
+}
```
john-z-yang, line 14: I think it might be worthwhile to just let the `RuntimeConfigManager` also take ownership of the green thread that polls the file. Something like
```
pub struct RuntimeConfigManager {
    handle: JoinHandle<()>
}
```

```rust
impl RuntimeConfigManager {
    pub fn new(path: String) -> Self {
        handle: tokio::spawn(async {
            // file IO here
        });
    }
```

```rust
impl Drop for RuntimeConfigManager {
    fn drop(&mut self) {
        self.handle.abort();
    }
}
```

src/runtime_config.rs
john-z-yang, line 29: Crashing due to a badly formatted config file might be a little extreme, maybe we can just emit an error here instead and implement a default for `RuntimeConfig`

src/main.rs
enochtangg: Updated `reload_config` to log when runtime config has changed and successfully reloaded. This also returns a result indicating whether the config has been reloaded.

src/runtime_config.rs
enochtangg: Yeah, that'd be quite catastrophic. Updated to log an error instead!

src/main.rs
enochtangg: If we eventually require the it to need runtime configs, then yes. Otherwise, dropping tasks via killswitching happens outside of `InflightActivationStore`

src/runtime_config.rs
enochtangg: Good catch, updated to use `tokio::fs`

src/runtime_config.rs
enochtangg: Agreed, added a default for the config and change this to emit an error instead.

src/runtime_config.rs
enochtangg: What is the advantage of scheduling the tokio thread inside `RuntimeConfigManager` rather than in main alongside the maintenance task? If we put this in the config manager, how do we call it periodically?

john-z-yang: For me the main advantage is that `RuntimeConfigManager ` is a self contained thing that doesn't rely on another thing to make it work properly. Right now, the contents of `RuntimeConfigManager` is only correct if something else it updating it.

> What is the advantage of scheduling the tokio thread inside `RuntimeConfigManager` rather than in main alongside the maintenance task? If we put this in the config manager, how do we call it periodically?

You don't need to call it periodically anymore, since it reloads on its own in the green thread it owns.

```rust
impl RuntimeConfigManager {
    pub fn new(path: String) -> Self {
        handle: tokio::spawn(async {
            loop {
                sleep(...).await;
                 // file IO here
            }
        });
    }
```

john-z-yang: > In your snippet, we'd sleep right? Is there a difference between sleeping and using interval.tick()? Do both yield execution to the scheduler?

Yup, both will yield execution to the scheduler because they're both using .await. I think it's probably better to use interval instead of sleep here.

The difference between interval and sleep is that sleep is yielding execution for the duration of its parameter, where as by default interval fires in walltime. For example
```rust
loop {
    a();
    sleep(Duration::from_secs(1).await;
}
```
vs
```rust
let timer = time::interval(Duration::from_secs(1);
loop {
    a();
    timer.tick().await;
}
```

In the former, it does not matter how long calling `a()` will take. Once a() finishes, we will sleep for 1 second. But in the second example, timer.tick().await will take until it has been 1 second since last time timer.tick().await has returned. This gets kind of complicated when `a()` takes longer than 1 second and if we configure the [MissedTickBehavior](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html).

john-z-yang: Also circling back to the original conversation, I think if we want to drive the update of `RuntimeConfig` explicitly in main, it's also fine since we're doing that with upkeep anyway.

src/config.rs
```diff
@@ -116,6 +119,7 @@ impl Default for Config {
             kafka_auto_offset_reset: "latest".to_owned(),
             kafka_send_timeout_ms: 500,
             db_path: "./taskbroker-inflight.sqlite".to_owned(),
+            runtime_config_path: "./runtime_config_defaults.yaml".to_owned(),
```
john-z-yang, line 122: Do we want this to be an Option<String> instead, so are we expecting that this file will always exists? If so, let's add it to the integration tests because the rebalancing one is failing from not being able to find this file

enochtangg: Right, now that we have defaults, we don't need to read from a default file anymore. Updated!

src/runtime_config.rs
```diff
@@ -0,0 +1,94 @@
...
+        }
+    }
+
+    async fn reload_config(path: &Option<String>, config: &Arc<RwLock<RuntimeConfig>>) {
+        if path.is_none() {
+            return;
+        }
+        let yaml_path = path.as_ref().unwrap();
+        let contents = fs::read_to_string(yaml_path).await;
+        let new_config: RuntimeConfig = match contents {
```
evanh, line 42: I think there is a case here that needs to be addressed:

- We successfully set some killswitches
- We then want to add something else to the list, but mess up the yaml

In that case, all the original changes will be overridden by the default config. In that case I would rather the server continue using the existing config, instead of the default.

src/runtime_config.rs
```diff
@@ -0,0 +1,94 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::RuntimeConfigManager;
+    use tokio::fs;
+
+    #[tokio::test]
+    async fn test_runtime_config_manager() {
+        let test_yaml = r#"
+drop_task_killswitch:
```
markstory, line 81: It would be good to have additional tests covering

- The config file missing.
- The config file containing invalid yaml.
 
As Evan mentioned we should preserve the existing configuration in both of these scenarios as we need to be resilient to bad configuration deploys.

src/runtime_config.rs
enochtangg: That's a good point. Updated to use existing runtime configs instead of overriding with defaults if yaml is malformed. 👍

src/runtime_config.rs
enochtangg: Addressed Evan's point and added 3 other tests: missing config test, invalid yaml test, and preserving existing configs test if reload sees either a missing or invalid yaml.

PR #257, opened 2025-03-16
Use join_all instead of JoinSet for upkeep

john-z-yang:


john-z-yang:
Doesn't seem to use less cpu, and uses way more memory

PR #258, opened 2025-03-17
feat(consumer): Add filtering to processing strategy and ability to killswitch tasks.

enochtangg:
### Overview
Taskbroker and taskworkers are susceptible to being overwhelmed by poorly performing or broken tasks. This PR is responsible for adding the ability to load shed by dropping matching tasks in the consumer according to killswitches defined on the runtime configuration. When a task has killswitch active the following should become true:
1. Any newly ingested activations are discarded. Activations will not be stored in Sqlite and will be unavailable to workers.
2. Any activations currently stored in sqlite with `status=pending` will be deleted, removing those tasks from the pool that workers can access. Processing tasks will be left unharmed as we should acknowledge the processing results of those tasks. **Note: This PR does not cover this change. This will be added in a subsequent change to avoid cluttering this PR.**

To drop tasks, the following runtime configuration is required: 
```
# /opt/taskbroker-runtime-config.yaml
drop_task_killswitch
    - "taskname-to-be-dropped"
```

To accomplish this, a filter step was added to the processing strategy after map and before reduce. Similar to the reduce step, the macro matches on 0->N number of generic filters, creates a separate green thread for each filter, and string them together with a mpsc channel. The diagram below indicates the new logical data flow of the taskbroker consumer 

image

```
processing_strategy!({
    map:
        deserialize_activation::new(),

    filter:
        DropActivationKillswitch::new(
            runtime_config.clone()
        ),

    reduce:
        InflightActivationBatcher::new(
            ActivationBatcherConfig::from_config(&consumer_config)
        )
        => InflightActivationWriter::new(
            consumer_store.clone(),
            ActivationWriterConfig::from_config(&consumer_config)
        ),

    err:
        OsStreamWriter::new(
            Duration::from_secs(1),
            OsStream::StdErr,
        ),
})
```

A custom filter called `DropActivationKillswitch` was created and receives the parsed InflightActivation from map as an input and compares the taskname against the runtime config. If the taskname exists in the killswitch list, then it will emit a counter and drop the task.

### Testing
- [x] Successfully ran end-to-end local tested with killswitch set in runtime config.
- [x] Unit testing
- [ ] TODO: Run a sandbox test and observe effects on performance.

### DIFF
--- a/src/kafka/consumer.rs
+++ b/src/kafka/consumer.rs
@@ -213,6 +213,53 @@ impl ActorHandles {
 #[macro_export]
 macro_rules! processing_strategy {
     (
+        @filters,
+        ($filter:expr),
+        $prev_receiver:ident,
+        $err_sender:ident,
+        $shutdown_signal:ident,
+        $handles:ident,
+    ) => {{
+        let (filter_sender, reduce_receiver) = tokio::sync::mpsc::channel(1);
+
+        $handles.spawn($crate::kafka::consumer::filter(
+            $filter,
+            $prev_receiver,
+            filter_sender.clone(),
+            $err_sender.clone(),
+            $shutdown_signal.clone(),
+        ));
+
+        (filter_sender, reduce_receiver)
+    }};
+    (
+        @filters,
+        ($filter_first:expr $(,$filter_rest:expr)+),
+        $prev_receiver:ident,
+        $err_sender:ident,
+        $shutdown_signal:ident,
+        $handles:ident,
+    ) => {{
+        let (sender, receiver) = tokio::sync::mpsc::channel(1);
+
+        $handles.spawn($crate::kafka::consumer::filter(
+            $filter_first,
+            $prev_receiver,
+            sender.clone(),
+            $err_sender.clone(),
+            $shutdown_signal.clone(),
+        ));
+
+        processing_strategy!(
+            @filters,
+            ($($filter_rest),+),
+            receiver,
+            $err_sender,
+            $shutdown_signal,
+            $handles,
+        )
+    }};
+    (
         @reducers,
         ($reduce:expr),
         $prev_receiver:ident,
@@ -262,6 +309,7 @@ macro_rules! processing_strategy {
     (
         {
             map: $map_fn:expr,
+            filter: $filter_first:expr $(=> $filter_rest:expr)*,
             reduce: $reduce_first:expr $(=> $reduce_rest:expr)*,
             err: $reduce_err:expr,
         }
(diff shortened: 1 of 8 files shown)


src/kafka/consumer.rs
```diff
@@ -489,6 +546,54 @@ pub async fn map<T>(
...
+                debug!("Receive shutdown signal, shutting down...");
+                break;
+            }
+
+            val = receiver.recv() => {
+                let Some((msgs, value)) = val else {
+                    break;
+                };
+                match filter.filter(&value).await {
+                    Ok(true) => {}
```
john-z-yang, line 576: This swallows the value when something is filtered. The offset will not be committed causing us to lose progress.

john-z-yang (changes requested):
I don't really see a very strong argument for a dedicated filter step to add to the consumer macro. There're a few reasons I am against it.

1. The filter needs to forward the offset of the filtered message to the subsequent steps, all the way to commit, otherwise it will not be committed on consumer shutdown
2. I don't really see the benefit of adding a filter step over having either: Change the map step to `Fn(Arc<OwnedMessage>) -> Result<Option<InflightActivation>, Error>` so it returns `None` if the messages are filtered, or just add a check to the reduce function to see if the activation belongs to a filtered namespaces when writing to the buffer of the InflightActivationBatcher [here](https://github.com/getsentry/taskbroker/blob/78cee0094cc68ac26c4deb4f2985158b7808d308/src/kafka/inflight_activation_batcher.rs#L43).
```rust
impl Reducer for InflightActivationBatcher {
    type Input = InflightActivation;

    type Output = Vec<InflightActivation>;

    async fn reduce(&mut self, t: Self::Input) -> Result<(), anyhow::Error> {
        // I think this is all you need
        if (!filtered_namespaces.containes(t.namespace)) {
            self.buffer.push(t);
        }
        Ok(())
    }
```

markstory:
Could we try filtering in the reduce step? That seems like it might be a simpler solution.

enochtangg:
> This swallows the value when something is filtered. The offset will not be committed causing us to lose progress.

Oh right, good catch. 

@john-z-yang @markstory
Okay, I'd also be inclined to do the more simple thing first and expand the processing strategy as we need. I suppose this feature might not doesn't warrant that yet. However, in the future, if we do require to do some more complex filtering, I think it'd be nice to move this there. For context, we have something similar that exists in snuba generic metrics consumers which uses a filter step in the arroyo processing strategy to drop specific messages based on certain headers.

enochtangg:
Closed in favour for: https://github.com/getsentry/taskbroker/pull/260

john-z-yang:
> > This swallows the value when something is filtered. The offset will not be committed causing us to lose progress.
> 
> Oh right, good catch.
> 
> @john-z-yang @markstory Okay, I'd also be inclined to do the more simple thing first and expand the processing strategy as we need. I suppose this feature might not doesn't warrant that yet. However, in the future, if we do require to do some more complex filtering, I think it'd be nice to move this there. For context, we have something similar that exists in snuba generic metrics consumers which uses a filter step in the arroyo processing strategy to drop specific messages based on certain headers.

Yes I agree, I think this is worth considering if we want to make very complex filtering steps. I think filtering is tricky to do right because it is a mapping from 1 → (0 or 1) and we have to be careful of the 1 → 0 cases. Arroyo uses some shitty hacks by making their filters periodically send a watermark downstream so it can be committed, hence my preference for >=1  → 1. transforms. (and if you squint hard enough at it for our use case, a Filter for a type `T` is really just a Mapping of T → Option<T>.

PR #259, opened 2025-03-17
fix: Improve performance of SetTaskStatus

markstory:
Use `RETURNING *` to avoid having to emit an additional SELECT query just for metrics calculations. This should help improve throughput as we're emitting 1 fewer query on each SetTaskStatus operation.


evanh (approved):
Thanks for this!

PR #260, opened 2025-03-18
feat(killswitch): Drop task due to killswitch in batcher

enochtangg:
This is simpler implementation of dropping tasks without the processing strategy change described in: https://github.com/getsentry/taskbroker/pull/258. Instead, this change is made on the reduce step specifically in the `InflightActivationBatcher`. If the task name is found in the runtime config, then do not push into the buffer and emit a metric.

### DIFF
--- a/src/kafka/inflight_activation_batcher.rs
+++ b/src/kafka/inflight_activation_batcher.rs
@@ -1,6 +1,9 @@
-use std::{mem::replace, time::Duration};
+use std::{mem::replace, sync::Arc, time::Duration};
 
-use crate::{config::Config, store::inflight_activation::InflightActivation};
+use crate::{
+    config::Config, runtime_config::RuntimeConfigManager,
+    store::inflight_activation::InflightActivation,
+};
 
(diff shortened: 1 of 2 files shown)


john-z-yang (approved):

src/kafka/inflight_activation_batcher.rs
```diff
@@ -40,7 +48,15 @@ impl Reducer for InflightActivationBatcher {
     type Output = Vec<InflightActivation>;
 
     async fn reduce(&mut self, t: Self::Input) -> Result<(), anyhow::Error> {
-        self.buffer.push(t);
+        let runtime_config = self.runtime_config_manager.read().await;
+        let task_name = &t.activation.taskname;
+        if !runtime_config.drop_task_killswitch.contains(task_name) {
```
evanh, line 53: I realize I should have thought of this on your previous PR, but it would be nice if this was a HashSet() instead of a Vec for these contains checks. In practice it's probably not a big deal, but just in case.

PR #261, opened 2025-03-18
feat: Allow fetching multiple tasks at once

evanh:
Change the store to allow fetching multiple pending tasks at once. This is the first step towards being able to batch fetch tasks from the worker.

This adds a `get_pending_activations` function, and points the existing function to call that function with a limit of 1. In the future new endpoints can leverage the pending activations function directly. That function accepts the namespaces that are being requested, and the number of pending tasks that can be returned.


markstory (commented):
Looks good to me. I don't think we'll need a batch API in the immediate future. We could hold onto this until the need for batch get is more evident though.

george-sentry:
We are moving away from the pull model and my recent changes have already implemented such a feature. Evan agrees that we can close this PR 🙂

PR #284, opened 2025-03-31
docker: don't run as root

keeakita:
Attempt to use a non-root user for the Docker image.


john-z-yang (approved):
This looks good to me, thank you so much!

PR #286, opened 2025-04-01
ref(docker): Stop using config files

john-z-yang:
Using a config file instead of the default config specified in code is a bit of a footgun.

### DIFF
--- a/config/config-sentry-dev.yaml
+++ b/config/config-sentry-dev.yaml
@@ -1,3 +1,0 @@
-upkeep_task_interval_ms: 2000
-kafka_cluster: 127.0.0.1:9092
-create_missing_topics: true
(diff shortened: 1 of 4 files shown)


config/config-sentry-dev.yaml
```diff
@@ -1,3 +0,0 @@
-upkeep_task_interval_ms: 2000
-kafka_cluster: 127.0.0.1:9092
-create_missing_topics: true
```
markstory, line 3: With these config files removed, we'll need another way to wire up taskbroker to the rest of the sentry dev environment. I think the current devservices setup relies on the container having a configuration file present.

john-z-yang: I agree, will have a discussion with dev infra sometime this week

PR #287, opened 2025-04-01
ref(kafka): continue to poll consumer on broker transport error

john-z-yang:
We stop polling the kafka client when we encounter an error, but this is not desirable as transport failure can result in a value yielded from the rdkafka client.
So if multiple brokers are supplied to the consumer, but a single broker is down, the consumer will crashloop.


evanh (approved):
Can we add a unit test for this?

john-z-yang:
> Can we add a unit test for this?

@evanh Not easily because this is part of the kafka library

PR #289, opened 2025-04-03
ref(config): Add producer configs

john-z-yang:


evanh (approved):

src/config.rs
```diff
@@ -54,6 +54,10 @@ pub struct Config {
     /// Whether to create missing topics if they don't exist.
     pub create_missing_topics: bool,
 
+    /// Comma separated list of kafka brokers to
+    /// publish dead letter messages on
+    pub kafka_deadletter_cluster: String,
```
markstory, line 59: Should this default to the value for `kafka_cluster` if it is undefined?

markstory (approved):

src/config.rs
john-z-yang: Yea we can do that

PR #291, opened 2025-04-03
ref(consumer): clean up processing_strategy macro

john-z-yang:
1. Changes the base case of the `@reducer` internal rule to 0 reducers from 1 reducers to make the base case much simpler
2. Changes the reducer chaining symbol from `=>` to `,` for aesthetics

Verified that the generated code is the same
https://www.diffchecker.com/Zgz3tIxc/

### DIFF
--- a/src/kafka/consumer.rs
+++ b/src/kafka/consumer.rs
@@ -223,27 +223,17 @@ impl ActorHandles {
 macro_rules! processing_strategy {
     (
         @reducers,
-        ($reduce:expr),
+        (),
         $prev_receiver:ident,
         $err_sender:ident,
         $shutdown_signal:ident,
         $handles:ident,
     ) => {{
-        let (commit_sender, commit_receiver) = tokio::sync::mpsc::channel(1);
-
-        $handles.spawn($crate::kafka::consumer::reduce(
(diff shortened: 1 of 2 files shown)


src/kafka/consumer.rs
```diff
@@ -223,27 +223,17 @@ impl ActorHandles {
...
-            $shutdown_signal.clone(),
-        ));
-
-        (commit_sender, commit_receiver)
+        $prev_receiver
     }};
     (
         @reducers,
-        ($reduce_first:expr $(,$reduce_rest:expr)+),
+        ($reduce_first:expr $(,$reduce_rest:expr)*),
```
markstory, line 236: I'm not well versed in macros, but is this the change that enabled 232-242 to be removed as the single reducer case isn't special anymore?

john-z-yang: Yes pretty much, this makes the base case 0 reducer, so the recursive case is 1 or more reducers. Before the base case is 1 reducer, so the recursive case has to be 2 or more, hence `reduce_rest` was one or more (aka +), not zero or more (aka *)

markstory (approved):

PR #297, opened 2025-04-07
wip(taskbroker): Add consumer step to resend delayed task to kafka

john-z-yang:
TODO:
- Make parameters configurable
- Separate our 3 kafka configs better (consumer config, producer for retry config, producer for delayed tasks config)
- Determine if there's a better retry behaviour when kafka producer fail to write delayed tasks.
- Add some tests for the 3 reducer steps

### DIFF
--- a/src/kafka/kafka_writer.rs
+++ b/src/kafka/kafka_writer.rs
@@ -1,0 +1,145 @@
+use super::consumer::{
+    ReduceConfig, ReduceShutdownBehaviour, ReduceShutdownCondition, Reducer,
+    ReducerWhenFullBehaviour,
+};
+use crate::{config::Config, store::inflight_activation::InflightActivation};
+use core::panic;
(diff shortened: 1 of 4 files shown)


src/kafka/kafka_writer.rs
```diff
@@ -0,0 +1,145 @@
...
+                        producer
+                            .send(
+                                FutureRecord::<(), Vec<u8>>::to(&config.kafka_topic)
+                                    .payload(&payload),
+                                Timeout::After(Duration::from_millis(config.kafka_send_timeout_ms)),
+                            )
+                            .await
+                    }
+                })
+                .collect::<FuturesUnordered<_>>();
```
markstory, line 98: We have this pattern in a few places now, should we try making a function for it?

john-z-yang: We should, they're all slightly different tho. So we need to be clever with our abstraction.

PR #309, opened 2025-04-17
feat(integration-test): Delayed task integration test

enochtangg:
This PR is responsible for testing the end-to-end flow of delayed tasks within taskbroker and adding this test as a required check in CI.

### What does this test do?
This test is responsible for checking the integrity of the countdown mechanism responsible for delaying tasks to be executed at a later time. This functionality is made possible by the the upkeep thread of taskbroker which is responsible for shifting tasks from the delay state to the pending state after the countdown has elapsed. In this integration test, an initial amount of messages is produced to kafka with a set delay time. These messages are first written into sqlite with a delay state. During an interval, the upkeep thread periodically checks whether a delayed tasks can be updated. This process continues until all tasks have been shifted to a pending state.

The test starts 1 broker thread. Once the upkeep thread has successfully shifted all delayed tasks in sqlite, taskbroker shuts down. Finally, this total number of delayed tasks and whether the delay has elapsed is validated.

### DIFF
--- a/python/integration_tests/test_upkeep_delay.py
+++ b/python/integration_tests/test_upkeep_delay.py
@@ -25,9 +25,7 @@ from sentry_protos.taskbroker.v1.taskbroker_pb2 import TaskActivation
 TEST_OUTPUT_PATH = TESTS_OUTPUT_ROOT / "test_upkeep_delay"
 
 
-def generate_task_activation(
-    delay: int
-) -> TaskActivation:
+def generate_task_activation(delay: int) -> TaskActivation:
     return TaskActivation(
         id=uuid4().hex,
         namespace="integration_tests",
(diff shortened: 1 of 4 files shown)


python/integration_tests/test_upkeep_delay.py
```diff
@@ -0,0 +1,292 @@
...
+
+        pending_count = task_count_in_sqlite.get("Pending", 0)
+        with open(results_log_path, "a") as results_log_file:
+            results_log_file.write(
+                f"total_delayed_tasks:{delay_count},"
+                f"total_pending_tasks:{pending_count},"
+                f"delay_has_elapsed:{int(cur_time >= end_of_delay)}"
+            )
+
+        time.sleep(5)  # Give some extra time for tasks to flush to DLQ topic
```
markstory, line 146: Is there a DLQ for this test?

python/integration_tests/test_upkeep_delay.py
```diff
@@ -0,0 +1,292 @@
...
+                finished_writing_tasks = True
+            time.sleep(1)
+
+        if not finished_writing_tasks:
+            print(
+                "[taskbroker_0]: Taskbroker was unable to write all tasks to "
+                "sqlite within timeout."
+            )
+
+        # Keep runniung taskbroekr until:
```
markstory, line 84:
```suggestion
        # Keep running taskbroker until:
```

markstory (approved):

evanh (approved):

PR #310, opened 2025-04-17
feat(reduce): Backpressure when delay task limit reached

enochtangg:
### Overview
Currently, if a very large influx of delayed tasks are produced to kafka, there is nothing stopping from the sqlite from growing unboundedly and running out of disk. This PR is responsible for adding a upper limit to the number of delayed tasks that can be stored in sqlite at a given moment. If an incoming batch exceeds this number, then the reduce step will stop receiving messages in its buffer, which bubbles up the pipeline and stops the consumer from reading more messages from kafka.

This has a few implications:
- Since we write in batches, we do not distinguish between delayed tasks and regular tasks at the reduce step.
- Therefore, if the incoming batch causes either the `max_pending_count` or `max_delay_count` to exceed, then emit backpressure. This means that either of these limits can halt ingestion despite the other limit not being reached.
- We should be careful with `max_delay_count`. Even though we have ample disk available, there can be scenarios where a bunch (or all) delayed tasks moved to pending at once. If `max_delay_count` is set to high, this would cause a large amount of task to move to the pending state which will halt ingestion until all pending tasks are burned below 2048 (our current default). I've set this default to `8192` for now.

Other change:
- Added a `upkeep.delay_count` gauge metric which shows the active number of delayed tasks sitting in sqlite at a given moment

### DIFF
--- a/src/kafka/inflight_activation_writer.rs
+++ b/src/kafka/inflight_activation_writer.rs
@@ -90,13 +90,12 @@ impl Reducer for InflightActivationWriter {
             .iter()
             .any(|activation| activation.status == InflightActivationStatus::Pending);
 
-        // We want to allow the batch to be inserted if:
-        // 1. The batch does not exceed both the pending and delay limits
-        // 2. The batch exceeds the pending limit, does not exceed delay limit, and is entirely delay
-        // 3. The batch exceeds the delay limit, does not exceed pending limit, and is entirely pending
-        // Otherwise, emit backpressure.
-        if exceeded_pending_limit && exceeded_delay_limit
-            || exceeded_delay_limit && has_delay
+        // Backpressure if any of these conditions are met:
+        // 1. The delay limit is exceeded AND either:
+        //    - there are delay activations in the batch, OR
+        //    - the pending limit is also exceeded
+        // 2. The pending limit is exceeded AND there are pending activations
+        if (has_delay || exceeded_pending_limit) && exceeded_delay_limit
             || exceeded_pending_limit && has_pending
         {
             return Ok(None);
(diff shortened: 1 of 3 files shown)


markstory (approved):

src/kafka/inflight_activation_writer.rs
```diff
@@ -73,6 +77,17 @@ impl Reducer for InflightActivationWriter {
             return Ok(None);
         }
 
+        if self
+            .store
+            .count_by_status(InflightActivationStatus::Delay)
+            .await
+            .expect("Error communicating with activation store")
+            + batch.len()
```
evanh, line 85: I don't think we want to limit purely by `batch.len` here. This could mean that a batch that contains no delay tasks will get blocked. I think the batch should only backpressure if it is actually adding more delay tasks.

evanh (commented):
Can we add a test?

src/kafka/inflight_activation_writer.rs
evanh: I'm more concerned about this case then the case of delayed tasks being blocked.

If we have to pass the numbers through when the batch is created then let's do that. I don't want it to be possible for ingestion tasks to stop because of some delayed task.

enochtangg: Okay I've updated this PR to scan for delay and pending tasks in the reduce step. Then, depending on what is currently sitting in sqlite, only write the activations that make the limit exceeded. This allows pending tasks to be written even when delayed task are at max capacity. I've also added some tests. I'll need a re-review on this

enochtangg:
As discussed offline, if only one of the limit (pending limit or delay limit) is reached, and the entire batch only contains tasks of the other status, then we still want to write into sqlite.

markstory (approved):

markstory (approved):

Issue #320, opened 2025-04-28
Add support for kafka sasl authentication to taskbroker


markstory:
To support the work that security is doing to enable authentication for kafka producer + consumer connections.

The environment variables are being added in: https://github.com/getsentry/ops/pull/15203

PR #321, opened 2025-04-28
feat: Add support for sasl authentication

markstory:
Add support for simple authentication to support work security team is doing.

Refs getsentry/ops#15203
Refs #320


evanh (commented):
It doesn't seem like we actually do anything with these values. Why is that?

markstory:
> It doesn't seem like we actually do anything with these values. Why is that?

The plan is for them to be set via environment variables as our kafkas will start requiring sasl credentials soon.

PR #322, opened 2025-04-29
feat: Add small single tenants to deploys

markstory:
Enable deployments for the small single tenants.


evanh (approved):
I'm not sure what this is doing, but the deploy pipeline already runs on these customers.

PR #336, opened 2025-05-05
feat(taskbroker): Drop expired tasks early

enochtangg:
This PR is responsible for dropping expired tasks earlier. When a task get consumed and reaches the reduce step, we now check whether the task is killswitched OR whether the task has already expired. If so, drop the task. By skipping the store + upkeep earlier, we would likely get some more throughput from the consumer when experiencing large backlogs of expired tasks. Additionally, this PR changes the way upkeep does `handle_expired_at()`. Instead of hopping the task from pending -> failure -> complete, just delete the activation immediately. This has two implications:
1. Taskbroker will not support DLQing expired tasks.
2. The expire metric will now indicate that the task has been purged from SQLite (previously, this metric only indicated that the activation has been set from pending to failure. This was usually coupled with the discarded and completed counter metrics).

### DIFF
--- a/src/kafka/inflight_activation_batcher.rs
+++ b/src/kafka/inflight_activation_batcher.rs
@@ -1,9 +1,9 @@
-use std::{mem::replace, sync::Arc, time::Duration};
-
 use crate::{
     config::Config, runtime_config::RuntimeConfigManager,
     store::inflight_activation::InflightActivation,
 };
+use chrono::Utc;
+use std::{mem::replace, sync::Arc, time::Duration};
 
 use super::consumer::{
     ReduceConfig, ReduceShutdownBehaviour, ReduceShutdownCondition, Reducer,
@@ -50,12 +50,22 @@ impl Reducer for InflightActivationBatcher {
     async fn reduce(&mut self, t: Self::Input) -> Result<(), anyhow::Error> {
         let runtime_config = self.runtime_config_manager.read().await;
         let task_name = &t.activation.taskname;
-        if !runtime_config.drop_task_killswitch.contains(task_name) {
-            self.buffer.push(t);
-        } else {
+        // drop the task if it has expired
+
+        if runtime_config.drop_task_killswitch.contains(task_name) {
             metrics::counter!("filter.drop_task_killswitch", "taskname" => task_name.clone())
                 .increment(1);
+            return Ok(());
         }
+
+        if let Some(expires_at) = t.expires_at {
+            if Utc::now() > expires_at {
+                metrics::counter!("upkeep.expired").increment(1);
+                return Ok(());
+            }
+        }
+
+        self.buffer.push(t);
 
         Ok(())
     }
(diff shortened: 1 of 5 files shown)


src/kafka/inflight_activation_batcher.rs
```diff
@@ -50,13 +50,22 @@ impl Reducer for InflightActivationBatcher {
...
+
+        if runtime_config.drop_task_killswitch.contains(task_name) {
             metrics::counter!("filter.drop_task_killswitch", "taskname" => task_name.clone())
                 .increment(1);
+            return Ok(());
         }
 
+        if let Some(expires_at) = t.expires_at {
+            if Utc::now() > expires_at {
+                metrics::counter!("upkeep.expired").increment(1);
```
evanh, line 62: I think we should call this `upkeep.expired_early` or something like that? So we can tell the difference?

enochtangg, line 62: Maybe this should have a different metric name now

src/kafka/inflight_activation_batcher.rs
```diff
@@ -50,13 +50,22 @@ impl Reducer for InflightActivationBatcher {
...
-            self.buffer.push(t);
-        } else {
+
+        if runtime_config.drop_task_killswitch.contains(task_name) {
             metrics::counter!("filter.drop_task_killswitch", "taskname" => task_name.clone())
                 .increment(1);
+            return Ok(());
         }
 
+        if let Some(expires_at) = t.expires_at {
```
evanh, line 60: Can you add a unit test for this too please.

src/store/inflight_activation.rs
```diff
@@ -563,19 +563,15 @@ impl InflightActivationStore {
     #[instrument(skip_all)]
     pub async fn handle_expires_at(&self) -> Result<u64, Error> {
         let now = Utc::now();
-        let update_result = sqlx::query(
-            r#"UPDATE inflight_taskactivations
-            SET status = $1
-            WHERE expires_at IS NOT NULL AND expires_at < $2 AND status = $3
-            "#,
+        let query = sqlx::query(
```
evanh, line 566: By doing this you will be throwing off our "completed" count metric. That's not a problem, but something to note.

src/kafka/inflight_activation_batcher.rs
enochtangg: Yes, added 👍

src/kafka/inflight_activation_batcher.rs
enochtangg: Renamed to `filter.expired_at_consumer` 🤔

evanh (approved):

src/store/inflight_activation.rs
```diff
@@ -563,19 +563,15 @@ impl InflightActivationStore {
...
+            "DELETE FROM inflight_taskactivations WHERE expires_at IS NOT NULL AND expires_at < $1",
```
markstory, line 567: Shouldn't we also keep the condition on status = Pending? There is a bind for it still as well.

PR #337, opened 2025-05-05
ref(taskworker): make upkeep metrics more ergonomic

john-z-yang:
Instead of creating a separate time series for each kind of task state and cleanup action. This pr changes them into a single time series with different tags, so that we just need to put a single time series when creating dashboard.

Also renames the gauge metrics to make them sound less confusing.

### DIFF
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
     metrics::histogram!("upkeep.duration").record(upkeep_start.elapsed());
 
     // Task statuses
-    metrics::counter!("upkeep.completed").increment(result_context.completed);
-    metrics::counter!("upkeep.failed").increment(result_context.failed);
-    metrics::counter!("upkeep.retried").increment(result_context.retried);
-    metrics::counter!("upkeep.expired").increment(result_context.expired);
+    metrics::counter!("upkeep.task.state_transition", "state" => "completed")
+        .increment(result_context.completed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "failed")
+        .increment(result_context.failed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "retried")
+        .increment(result_context.retried);
+    metrics::counter!("upkeep.task.state_transition", "state" => "expired")
+        .increment(result_context.expired);
 
     // Upkeep cleanup actions
-    metrics::counter!("upkeep.deadlettered").increment(result_context.deadlettered);
-    metrics::counter!("upkeep.discarded").increment(result_context.discarded);
-    metrics::counter!("upkeep.processing_attempts_exceeded")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "deadlettered")
+        .increment(result_context.deadlettered);
+    metrics::counter!("upkeep.cleanup_action", "kind" => "discarded")
+        .increment(result_context.discarded);
+    metrics::counter!("upkeep.cleanup_action", "kind" => "processing_attempts_exceeded")
         .increment(result_context.processing_attempts_exceeded);
-    metrics::counter!("upkeep.processing_deadline_reset")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "processing_deadline_reset")
         .increment(result_context.processing_deadline_reset);
-    metrics::counter!("upkeep.delay_elapsed").increment(result_context.delay_elapsed);
+    metrics::counter!("upkeep.cleanup_action", "kind" => "delay_elapsed")
+        .increment(result_context.delay_elapsed);
 
     // State of inflight tasks
-    metrics::gauge!("upkeep.pending_count").set(result_context.pending);
-    metrics::gauge!("upkeep.processing_count").set(result_context.processing);
-    metrics::gauge!("upkeep.delay_count").set(result_context.delay);
+    metrics::gauge!("upkeep.cur_pending_tasks").set(result_context.pending);
+    metrics::gauge!("upkeep.cur_processing_tasks").set(result_context.processing);
+    metrics::gauge!("upkeep.cur_delayed_tasks").set(result_context.delay);
 
     result_context
 }
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -295,9 +295,9 @@ pub async fn do_upkeep(
         .increment(result_context.delay_elapsed);
 
     // State of inflight tasks
-    metrics::gauge!("upkeep.cur_pending_tasks").set(result_context.pending);
-    metrics::gauge!("upkeep.cur_processing_tasks").set(result_context.processing);
-    metrics::gauge!("upkeep.cur_delayed_tasks").set(result_context.delay);
+    metrics::gauge!("upkeep.current_pending_tasks").set(result_context.pending);
+    metrics::gauge!("upkeep.current_processing_tasks").set(result_context.processing);
+    metrics::gauge!("upkeep.current_delayed_tasks").set(result_context.delay);
 
     result_context
 }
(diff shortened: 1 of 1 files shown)


src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
     metrics::histogram!("upkeep.duration").record(upkeep_start.elapsed());
 
     // Task statuses
-    metrics::counter!("upkeep.completed").increment(result_context.completed);
-    metrics::counter!("upkeep.failed").increment(result_context.failed);
-    metrics::counter!("upkeep.retried").increment(result_context.retried);
-    metrics::counter!("upkeep.expired").increment(result_context.expired);
+    metrics::counter!("upkeep.task.state_transition", "state" => "completed")
```
evanh, line 276: I'd like to have a metric here that captures the "success" state: tasks that didn't error/time out etc. Right now the "completed" metric just means that it was cleaned up out of upkeep. I think we should either add a new metric for the succeeding tasks or change completed to be that metric.

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
+    metrics::counter!("upkeep.cleanup_action", "kind" => "delay_elapsed")
+        .increment(result_context.delay_elapsed);
 
     // State of inflight tasks
-    metrics::gauge!("upkeep.pending_count").set(result_context.pending);
-    metrics::gauge!("upkeep.processing_count").set(result_context.processing);
-    metrics::gauge!("upkeep.delay_count").set(result_context.delay);
+    metrics::gauge!("upkeep.cur_pending_tasks").set(result_context.pending);
+    metrics::gauge!("upkeep.cur_processing_tasks").set(result_context.processing);
+    metrics::gauge!("upkeep.cur_delayed_tasks").set(result_context.delay);
```
markstory, line 300:
```suggestion
    metrics::gauge!("upkeep.current_pending_tasks").set(result_context.pending);
    metrics::gauge!("upkeep.current_processing_tasks").set(result_context.processing);
    metrics::gauge!("upkeep.current_delayed_tasks").set(result_context.delay);
```

We should use the full word imo.

markstory (approved):

src/upkeep.rs
john-z-yang: That sounds good to me, I think it can be a second set of tags to describe how the task is transitioned into complete.
```rust
metrics::counter!("upkeep.task.state_transition", "state" => "completed", "how" => "exec_successful")
metrics::counter!("upkeep.task.state_transition", "state" => "completed", "how" => "exec_failed")
```
Does this sound like what you had in mind?

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
-    metrics::counter!("upkeep.failed").increment(result_context.failed);
-    metrics::counter!("upkeep.retried").increment(result_context.retried);
-    metrics::counter!("upkeep.expired").increment(result_context.expired);
+    metrics::counter!("upkeep.task.state_transition", "state" => "completed")
+        .increment(result_context.completed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "failed")
+        .increment(result_context.failed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "retried")
+        .increment(result_context.retried);
+    metrics::counter!("upkeep.task.state_transition", "state" => "expired")
```
enochtangg, line 282: I recently changed how the expired metric behaves in: https://github.com/getsentry/taskbroker/pull/336. We should still have this metric, but now it makes more sense to have this as part of `upkeep.cleanup_action` instead since this now directly removes expired tasks from sqlite.

john-z-yang: Ah I see, done. Thanks!

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
+    metrics::counter!("upkeep.task.state_transition", "state" => "failed")
+        .increment(result_context.failed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "retried")
+        .increment(result_context.retried);
 
     // Upkeep cleanup actions
-    metrics::counter!("upkeep.deadlettered").increment(result_context.deadlettered);
-    metrics::counter!("upkeep.discarded").increment(result_context.discarded);
-    metrics::counter!("upkeep.processing_attempts_exceeded")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "deadlettered")
```
john-z-yang, line 284: I think preferably I like to rename these to something more descriptive, like `send_deadlettered_to_kafka` instead of `deadlettered`, or `remove_expired` instead of just `expired`.

src/upkeep.rs
evanh: No, I think I'd like a separate state, similar to how "failed' and "retried" are separate states.

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
+    metrics::counter!("upkeep.task.state_transition", "state" => "failed")
+        .increment(result_context.failed);
+    metrics::counter!("upkeep.task.state_transition", "state" => "retried")
+        .increment(result_context.retried);
 
     // Upkeep cleanup actions
-    metrics::counter!("upkeep.deadlettered").increment(result_context.deadlettered);
-    metrics::counter!("upkeep.discarded").increment(result_context.discarded);
-    metrics::counter!("upkeep.processing_attempts_exceeded")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "publish_deadlettered")
```
evanh, line 284: These names look good to me!

src/upkeep.rs
john-z-yang: Okay, but they're transitioning into the same state in the end (to complete) though. Wouldn't having them as separate states be confusing?

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
+        .increment(result_context.deadlettered);
+    metrics::counter!("upkeep.task.cleanup_action", "kind" => "mark_expired_as_failure")
+        .increment(result_context.expired);
+    metrics::counter!("upkeep.cleanup_action", "kind" => "delete_discarded")
+        .increment(result_context.discarded);
+    metrics::counter!("upkeep.cleanup_action", "kind" => "mark_processing_attempts_exceeded_as_failure")
         .increment(result_context.processing_attempts_exceeded);
-    metrics::counter!("upkeep.processing_deadline_reset")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "mark_processing_deadline_exceeded_as_failure")
```
markstory, line 292:
```suggestion
    metrics::counter!("upkeep.cleanup_action", "kind" => "processing_deadline_reset")
```

This should have a different kind value from line 290-291

markstory (commented):
The names look good to me outside of the one duplicate.

src/upkeep.rs
john-z-yang: One is processing attempt and another is processing deadline

src/upkeep.rs
```diff
@@ -273,24 +273,31 @@ pub async fn do_upkeep(
...
+    metrics::counter!("upkeep.task.state_transition", "state" => "retried")
+        .increment(result_context.retried);
 
     // Upkeep cleanup actions
-    metrics::counter!("upkeep.deadlettered").increment(result_context.deadlettered);
-    metrics::counter!("upkeep.discarded").increment(result_context.discarded);
-    metrics::counter!("upkeep.processing_attempts_exceeded")
+    metrics::counter!("upkeep.cleanup_action", "kind" => "publish_deadlettered")
+        .increment(result_context.deadlettered);
+    metrics::counter!("upkeep.task.cleanup_action", "kind" => "mark_expired_as_failure")
```
enochtangg, line 286: Names make sense to me, this one should more accurately be "remove_expired" or something like that? Since now we are deleting it from sqlite directly.

john-z-yang: Ah right, good call!

PR #340, opened 2025-05-05
fix(metric): Account for delayed tasks in latency metric

enochtangg:
Currently, our latency metric can incorrectly account for a task's delayed time. Since the latency metric is the duration between Kafka receive time and the time at which the activation was given to worker, delayed tasks can incorrectly skew this metric. Before emitting the metric, if a task was delayed, subtract the duration that it was delayed for, then emit its latency.

### DIFF
--- a/src/grpc/server.rs
+++ b/src/grpc/server.rs
@@ -8,9 +8,7 @@ use std::sync::Arc;
 use std::time::Instant;
 use tonic::{Request, Response, Status};
 
-use crate::store::inflight_activation::{
-    InflightActivation, InflightActivationStatus, InflightActivationStore,
-};
+use crate::store::inflight_activation::{InflightActivationStatus, InflightActivationStore};
 use tracing::{error, instrument};
 
 pub struct TaskbrokerServer {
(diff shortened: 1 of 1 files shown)


markstory (approved):

src/grpc/server.rs
```diff
@@ -37,16 +35,22 @@ impl ConsumerService for TaskbrokerServer {
...
                 ) {
+                    let mut received_to_gettask_latency = Utc::now()
+                        .signed_duration_since(sentry_ts)
+                        .num_milliseconds()
+                        as f64;
+                    if let Some(delay_until) = inflight.delay_until {
+                        received_to_gettask_latency -= delay_until
+                            .signed_duration_since(sentry_ts)
+                            .num_milliseconds()
+                            as f64;
```
markstory, line 46: This feels like we could have a method on `InflightActivation` that does this logic. Having a method would help reduce duplication.

john-z-yang, line 46: Might be a little more readable to use map_or here instead of mutation.
```rust
    let received_to_gettask_latency = Utc::now()
        .signed_duration_since(sentry_ts)
        .num_milliseconds()
        - inflight.delay_until.map_or(0, |delay_until| {
            delay_until
                .signed_duration_since(sentry_ts)
                .num_milliseconds()
        });
```

john-z-yang (approved):

PR #344, opened 2025-05-05
ref(dlq): Discard messages that fail to go to the DLQ

evanh:
If a message can't be written to the DLQ, instead of holding on to it indefinitely log the message out and discard the message.

### DIFF
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -326,7 +326,6 @@ mod tests {
     use std::time::Duration;
 
     use crate::{
-        config::Config,
         store::inflight_activation::{
             InflightActivationStatus, InflightActivationStore, InflightActivationStoreConfig,
         },
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -3,20 +3,23 @@ use futures::{StreamExt, stream::FuturesUnordered};
 use prost::Message;
 use prost_types::Timestamp;
 use rdkafka::{
-    message::OwnedMessage, producer::{FutureProducer, FutureRecord}, util::Timeout
+    producer::{FutureProducer, FutureRecord},
+    util::Timeout,
 };
 use sentry_protos::taskbroker::v1::TaskActivation;
 use std::{
-    fmt::Error, sync::Arc, time::{Duration, Instant}
+    sync::Arc,
+    time::{Duration, Instant},
 };
 use tokio::{select, time};
 use tracing::{error, info, instrument};
 use uuid::Uuid;
 
 use crate::{
-    config::Config, kafka::consumer::KafkaMessage, store::inflight_activation::{
+    config::Config,
+    store::inflight_activation::{
         InflightActivation, InflightActivationStatus, InflightActivationStore,
-    }
+    },
 };
 
 /// The upkeep task that periodically performs upkeep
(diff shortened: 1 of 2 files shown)


src/upkeep.rs
```diff
@@ -205,7 +205,13 @@ pub async fn do_upkeep(
 
                     match delivery {
                         Ok(_) => Ok(activation.id),
-                        Err(err) => Err(err),
+                        Err((err, _msg)) => {
+                            error!(
+                                "deadletter.publish.failure: {}, message: {:?}",
+                                err, payload
+                            );
+                            Err(err)
```
markstory, line 213: Should we be returning `Some(activation.id)` so that the activation will be deleted?

markstory (commented):
It would be good to have a test for this. We should be able to generate an oversized message that will fail.

src/upkeep.rs
```diff
@@ -593,6 +597,49 @@ mod tests {
...
+
+        let result_context = do_upkeep(config.clone(), store.clone(), producer).await;
+
+        // Only 1 record left as the failure task should be appended to dlq
+        assert_eq!(result_context.deadlettered, 1);
+        assert_eq!(store.count().await.unwrap(), 1);
+
+        let messages =
+            consume_topic(config.clone(), config.kafka_deadletter_topic.as_ref(), 1).await;
+        assert_eq!(messages.len(), 1);
```
markstory, line 634: If publishing to the DLQ fails wouldn't there be 0 messages in it?

src/upkeep.rs
```diff
@@ -218,10 +224,7 @@ pub async fn do_upkeep(
             .into_iter()
             .filter_map(|result| match result {
                 Ok(id) => Some(id),
-                Err((err, _msg)) => {
-                    error!("deadletter.publish.failure {}", err);
-                    None
-                }
+                Err((_err, _msg, id)) => Some(id),
```
john-z-yang, line 227: We don't need `filter_map` here because we're using using all of the ids from `deadletters`

src/upkeep.rs
```diff
@@ -205,7 +205,13 @@ pub async fn do_upkeep(
...
+                        Err((err, _msg)) => {
+                            error!(
+                                "deadletter.publish.failure: {}, message: {:?}",
+                                err, payload
+                            );
+                            Err((err, _msg, activation.id))
+                        }
                     }
                 }
             })
```
john-z-yang, line 217: Now that we're marking all activations that we're suppose to dead letter as complete, instead of only the ones we successfully sent to kafka. I don't think you need this to evaluate into `Ok(activation.id)` or `Err(activation.id, ...)`, you can just return `activation.id`
```rust
                async move {
                    ...
                    if let Err((err, _msg)) = delivery {
                        error!(
                            "deadletter.publish.failure: {}, message: {:?}",
                            err, payload
                        );
                    }
                    activation.id
                }
            })
           .collect::<FuturesUnordered<_>>();

evanh:
> It would be good to have a test for this. We should be able to generate an oversized message that will fail.

I can't figure out a way to test this that works both locally and in CI.

- Using a random topic for DLQ doesn't work, the producer will automatically create the topic
- Assigning a very low timeout doesn't work, the test runs too fast
- Creating a huge message doesn't work, either the message is too small to trigger an error or the test takes too long to run

markstory (approved):

john-z-yang (approved):

PR #345, opened 2025-05-05
fix(dlq): Add a way to override the max.messages.bytes in Kafka

evanh:
This setting controls the maximum size that the producer will send to Kafka.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -134,6 +134,10 @@ pub struct Config {
     /// The maximum number of seconds a task can be delayed until.
     /// Tasks delayed greater than this duration are capped.
     pub max_delayed_task_allowed_sec: u64,
+
+    /// The maximum size allowed for a message on the Kafka producer.
+    /// If a message is bigger than this then the produce will fail.
+    pub max_message_size: u64,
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -177,6 +181,7 @@ impl Default for Config {
             upkeep_task_interval_ms: 1000,
             maintenance_task_interval_ms: 6000,
             max_delayed_task_allowed_sec: 3600,
+            max_message_size: 10485760,
```
john-z-yang, line 184: How did we get this number? Using the DD metric?

enochtangg (approved):

src/config.rs
markstory: Should we use `10000000`? That's what we have in the topic schema configuration:

https://github.com/getsentry/sentry-kafka-schemas/blob/620bf410cdfd67497cc70f468a14a5c7189e24df/topics/taskworker.yaml#L16-L20

john-z-yang: Okay, I'm not opinionated on what it should be. I was just curious!

PR #350, opened 2025-05-08
feat(taskbroker): Add delay on retry

john-z-yang:
Bumps the proto version to 0.2.0, which adds the new fields to allow tasks to define delay upon a retry.
During upkeep, upon cloning/serializing a failed task for retry, we set the `delay` field if an activation has a `delay_on_retry` set.

Also adds a few unit tests as I forgot about this cloning/serialization part when implementing the original retry mechanism.

Worker changes
https://github.com/getsentry/sentry/pull/91197

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -732,12 +732,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "fixedbitset"
-version = "0.4.2"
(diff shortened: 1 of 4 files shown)


enochtangg (approved):

evanh (approved):
LGTM!

PR #355, opened 2025-05-09
fix(kafka): crash when consumer fails to write to sqlite

john-z-yang:


markstory (approved):

markstory:
Thinking about this over the weekend, should we backpressure instead of panic? If we backpressure to kafka when sqlite is overwhelmed/then grpc stays online, and workers will be able to complete any work they have in-flight, and we could drain sqlite to recover from it being overloaded.

PR #358, opened 2025-05-12
fix(upkeep): Add a semaphore to prevent concurrent upkeep/vacuum

evanh:
In a situtation where the broker is overloaded, and the upkeep/vacuum are taking longer than a single interval to run, the old functionality would simply continue calling the function. If the upkeep was taking 30 seconds to run, and there was a 1 second interval, that means there would be 30 concurrent upkeeps all running at the same time, and all effectively doing the same thing.

In the case of an incident like INC-1158, this would compound the issue the broker was facing: it was already under high load, and this would simply keep piling on more and more load, as well as potentially consuming a lot of memory keeping all the stack traces in memory.

Add a simple semaphore to limit the number of concurrent upkeeps/vacuums that can be running at once to avoid this compounding issue. Hopefully this will help prevent high load causing incidents.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -6,6 +6,7 @@ use taskbroker::kafka::inflight_activation_batcher::{
 };
 use taskbroker::upkeep::upkeep;
 use tokio::signal::unix::SignalKind;
+use tokio::sync::Semaphore;
 use tokio::task::JoinHandle;
 use tokio::{select, time};
 use tonic::transport::Server;
@@ -94,12 +95,16 @@ async fn main() -> Result<(), Error> {
         let maintenance_store = store.clone();
         let mut timer = time::interval(Duration::from_millis(config.maintenance_task_interval_ms));
         timer.set_missed_tick_behavior(time::MissedTickBehavior::Skip);
-
+        let max_vacuum = config.max_concurrent_vacuum;
         async move {
+            let vacuum_semaphore = Semaphore::new(max_vacuum);
             loop {
                 select! {
                     _ = timer.tick() => {
-                        let _ = maintenance_store.vacuum_db().await;
+                        let permit_attempt = vacuum_semaphore.try_acquire();
+                        if permit_attempt.err().is_none() {
+                            let _ = maintenance_store.vacuum_db().await;
+                        }
                         debug!("ran maintenance vacuum");
                     },
                     _ = guard.wait() => {
--- a/src/main.rs
+++ b/src/main.rs
@@ -102,7 +102,7 @@ async fn main() -> Result<(), Error> {
                 select! {
                     _ = timer.tick() => {
                         let permit_attempt = vacuum_semaphore.try_acquire();
-                        if permit_attempt.err().is_none() {
+                        if permit_attempt.is_ok() {
                             let _ = maintenance_store.vacuum_db().await;
                         }
                         debug!("ran maintenance vacuum");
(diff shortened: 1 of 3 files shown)


src/main.rs
```diff
@@ -94,12 +95,16 @@ async fn main() -> Result<(), Error> {
...
-
+        let max_vacuum = config.max_concurrent_vacuum;
         async move {
+            let vacuum_semaphore = Semaphore::new(max_vacuum);
             loop {
                 select! {
                     _ = timer.tick() => {
-                        let _ = maintenance_store.vacuum_db().await;
+                        let permit_attempt = vacuum_semaphore.try_acquire();
+                        if permit_attempt.err().is_none() {
```
markstory, line 105:
```suggestion
                        if permit_attempt.is_ok() {
```

We could check that the result is ok 🤷

src/upkeep.rs
```diff
@@ -32,13 +33,17 @@ pub async fn upkeep(config: Arc<Config>, store: Arc<InflightActivationStore>) {
...
+    let upkeep_semaphore = Semaphore::new(config.max_concurrent_upkeep);
     let guard = elegant_departure::get_shutdown_guard().shutdown_on_drop();
     let mut timer = time::interval(Duration::from_millis(config.upkeep_task_interval_ms));
     timer.set_missed_tick_behavior(time::MissedTickBehavior::Delay);
     loop {
         select! {
             _ = timer.tick() => {
-                let _ = do_upkeep(config.clone(), store.clone(), producer.clone()).await;
+                let permit_attempt = upkeep_semaphore.try_acquire();
+                if permit_attempt.err().is_none() {
```
markstory, line 44:
```suggestion
                if permit_attempt.is_ok() {
```

markstory (approved):

enochtangg (approved):

john-z-yang (changes requested):
> In a situtation where the broker is overloaded, and the upkeep/vacuum are taking longer than a single interval to run, the old functionality would simply continue calling the function. If the upkeep was taking 30 seconds to run, and there was a 1 second interval, that means there would be 30 concurrent upkeeps all running at the same time, and all effectively doing the same thing.

I don't think this happens. Because we're calling with `.await`?
```rust
do_upkeep(config.clone(), store.clone(), producer.clone()).await;
```
inside the loop, so we never go back to the next iteration of the loop until the `do_upkeep` function is done. Moreover, for upkeep's case, we have already set the [MissTickBehavior](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html) to [Delay](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Delay). So if the timer is 1 second, and upkeep takes 30 seconds, we will run the next upkeep 31 seconds after the first upkeep is called.

evanh:
> Which blocks until the function is done.

Reading the code I think you're right. I initially got put onto this track because we were seeing in the logs that the `upkeep.complete` log was happening ~200ms apart sometimes.

john-z-yang:
> > Which blocks until the function is done.
> 
> Reading the code I think you're right. I initially got put onto this track because we were seeing in the logs that the `upkeep.complete` log was happening ~200ms apart sometimes.

I do think we should align the [MissTickBehavior](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html) of the 2 timers, one of them is set to skip while the other one is delay

evanh:
This doesn't work the way I expected, closing.

PR #363, opened 2025-05-12
ref(kafka): retry sqlite write on failure

john-z-yang:


src/kafka/inflight_activation_writer.rs
```diff
@@ -101,36 +100,44 @@ impl Reducer for InflightActivationWriter {
...
+                    .increment(res.rows_affected);
+                debug!(
+                    "Inserted {:?} entries with max lag: {:?}s",
+                    res.rows_affected,
+                    lag.num_seconds()
+                );
+                Ok(Some(()))
+            }
+            Err(err) => {
+                error!("Unable to write to sqlite: {}", err);
```
evanh, line 137: I don't think this is necessary, but did you consider adding a sleep here?

evanh (approved):

src/kafka/inflight_activation_writer.rs
john-z-yang: Mmm yeah that's a good point, I will add a sleep here.

PR #369, opened 2025-05-15
evanh/feat/separate activations into another table

evanh:
- **get inflight activation store**
- **add migrations**
- **feat(sqlite): Split the activations into two tables**


migrations/0001_create_inflight_taskactivations.sql
```diff
@@ -6,7 +6,6 @@ CREATE TABLE IF NOT EXISTS inflight_taskactivations (
     added_at INTEGER NOT NULL,
     processing_attempts INTEGER NOT NULL,
     expires_at INTEGER,
-    processing_deadline_duration INTEGER NOT NULL,
```
john-z-yang, line 9: I don't think you can remove migrations. Since we already have databases in production with migrations already ran

src/config.rs
```diff
@@ -182,6 +188,8 @@ impl Default for Config {
             maintenance_task_interval_ms: 6000,
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 10485760,
+            blob_db_path: "./taskbroker-activation-blob.sqlite".to_owned(),
+            blob_delete_delay_ms: 30000,
```
markstory, line 192: Should we wait longer than 30s, some tasks take a few minutes to run.

src/store/inflight_activation.rs
```diff
@@ -74,10 +81,50 @@ impl From<TaskActivationStatus> for InflightActivationStatus {
...
+}
+
 #[derive(Clone, Debug, PartialEq)]
 pub struct InflightActivation {
+    /// The unique id of the activation
+    pub id: String,
+
     /// The protobuf activation that was received from kafka
-    pub activation: TaskActivation,
+    pub activation: Option<TaskActivation>,
```
markstory, line 127: I wonder if we should have two structs, one used during ingest to insert, and the other used by the store when reading/updating activation state.

src/store/inflight_activation.rs
```diff
@@ -131,6 +186,11 @@ impl InflightActivation {
                 })
         })
     }
+
+    /// A convenience function to get the TaskActivation
+    pub fn get_activation(&self) -> TaskActivation {
+        self.activation.clone().unwrap()
```
markstory, line 192: This seems like a risky method to call outside of tests. Should we mention that in the doc string?

src/store/inflight_activation.rs
```diff
@@ -280,21 +387,95 @@ impl InflightActivationStore {
...
+            SELECT activation 
+            FROM taskactivation_blobs 
+            WHERE id = $1
+            ",
+        )
+        .bind(id)
+        .fetch_optional(&self.blob_read_pool)
+        .await?;
+
+        let blob_row = blob_result.unwrap();
```
markstory, line 418: Do we want a panic on a missed id lookup?

src/store/inflight_activation.rs
```diff
@@ -280,21 +387,95 @@ impl InflightActivationStore {
...
 
     #[instrument(skip_all)]
     pub async fn store(&self, batch: Vec<InflightActivation>) -> Result<QueryResult, Error> {
         if batch.is_empty() {
             return Ok(QueryResult { rows_affected: 0 });
         }
+
+        // A note about the consistency of the data in the two databases:
+        // If there is data in the blob database that isn't in the meta database, that is not
+        // an inconsistency issue. No extra tasks will get run or failed. The orphaned blobs
```
markstory, line 438:
```suggestion
        // If there is data in the blob database that isn't in the meta database, that is not
        // a consistency issue. No extra tasks will get run or failed. The orphaned blobs
```

src/store/inflight_activation.rs
```diff
@@ -280,21 +387,95 @@ impl InflightActivationStore {
...
+        // will get cleaned up. However if there is data in the meta database that is not reflected
+        // in the blob database, that is an inconsistency error. As such, write the blobs first and
```
markstory, line 440:
```suggestion
        // in the blob database, that is a consistency error. As such, write the blobs first and
```

src/store/inflight_activation.rs
```diff
@@ -280,21 +387,95 @@ impl InflightActivationStore {
...
+            .push_values(blob_rows, |mut b, row| {
+                b.push_bind(row.id);
+                b.push_bind(row.activation);
+                b.push_bind(row.added_at.timestamp());
+            })
+            .push(" ON CONFLICT(id) DO NOTHING")
+            .build();
+        blob_query.execute(&self.blob_write_pool).await?;
+
+        // Sync the WAL into the main database so we don't lose data on host failure.
```
markstory, line 469:
```suggestion
        // Sync the WAL into the blob database so we don't lose data on host failure.
```

src/store/inflight_activation.rs
```diff
@@ -317,12 +499,12 @@ impl InflightActivationStore {
...
-            .collect::<Result<Vec<TableRow>, _>>()?;
+            .map(MetaTableRow::try_from)
+            .collect::<Result<Vec<MetaTableRow>, _>>()?;
         let query = query_builder
             .push_values(rows, |mut b, row| {
                 b.push_bind(row.id);
-                b.push_bind(row.activation);
+                b.push_bind(vec![]);
                 b.push_bind(row.partition);
                 b.push_bind(row.offset);
```
markstory, line 509: It will be good to remove these columns once we've tested this out further.

src/store/inflight_activation.rs
```diff
@@ -463,8 +647,14 @@ impl InflightActivationStore {
     pub async fn delete_activation(&self, id: &str) -> Result<(), Error> {
         sqlx::query("DELETE FROM inflight_taskactivations WHERE id = $1")
             .bind(id)
-            .execute(&self.write_pool)
+            .execute(&self.meta_write_pool)
+            .await?;
+
+        sqlx::query("DELETE FROM taskactivations WHERE id = $1")
```
markstory, line 653:
```suggestion
        sqlx::query("DELETE FROM taskactivation_blobs WHERE id = $1")
```

src/store/inflight_activation.rs
```diff
@@ -627,39 +821,37 @@ impl InflightActivationStore {
...
-            }
             // We could be deadlettering because of activation.expires
             // when a task expires we still deadletter if configured.
-            let retry_state = &activation.retry_state.as_ref().unwrap();
-            if retry_state.on_attempts_exceeded == OnAttemptsExceeded::Discard as i32
-                || retry_state.on_attempts_exceeded == OnAttemptsExceeded::Unspecified as i32
+            // let retry_state = &activation.retry_state.as_ref().unwrap();
+            let id: &str = record.get("id");
+            let infl_on_attempts_exceeded: InflightOnAttemptsExceeded =
+                record.get("on_attempts_exceeded");
```
markstory, line 846: Nice :clap: even if the separate blob db doesn't work out we should make this change so that we scan/load fewer bytes.

src/store/inflight_activation.rs
```diff
@@ -709,9 +901,38 @@ impl InflightActivationStore {
...
+            .execute(&self.meta_write_pool)
             .await?;
 
         Ok(query.rows_affected())
     }
+
+    /// Remove orphaned blobs that don't have a corresponding ID in the meta table.
+    /// This method is a garbage collector for the blob store.
+    #[instrument(skip_all)]
+    pub async fn remove_orphaned_blobs(&self) -> Result<u64, Error> {
```
markstory, line 913: We could call this from the maintenance thread.

src/store/inflight_activation_tests.rs
```diff
@@ -385,7 +400,7 @@ async fn test_handle_processing_deadline() {
     assert_count_by_status(&store, InflightActivationStatus::Processing, 0).await;
     assert_count_by_status(&store, InflightActivationStatus::Pending, 2).await;
 
-    let task = store.get_by_id(&batch[1].activation.id).await;
+    let task = store.get_by_id(&batch[1].id, false).await;
```
markstory, line 403: Should we have `get_by_id` and `get_by_id_with_activation`? 🤷

migrations/0001_create_inflight_taskactivations.sql
evanh: Here is what I was thinking:
Have four "modes" for how we store and fetch the activation blobs:
1. Write only to original table, read from original table
2. Dual write to both tables, read from original table
3. Dual write to both tables, read from new table
4. Write only to new table, read from the new table.

Then at any point we can go back to step 1 if we need. However we won't be able to go back before step 1 easily, because at that point the migration will have run to add the new columns to the original table, and the code will stop working. Using these modes, we can deploy the changes without them being active, then slowly turn them on to ensure everything works.

src/config.rs
evanh: This delay is only for orphaned IDs. So we only delete data that is orphaned from the meta table, and more delay_ms old.

src/grpc/server.rs
```diff
@@ -31,19 +33,28 @@ impl ConsumerService for TaskbrokerServer {
...
+                    Ok(activation) => activation,
+                    Err(e) => {
+                        error!("Unable to retrieve pending activation: {:?}", e);
+                        return Err(Status::internal("Unable to retrieve pending activation"));
+                    }
+                };
                 let now = Utc::now();
-                let received_to_gettask_latency = inflight.received_latency(now);
+                let received_to_gettask_latency =
+                    received_latency(activation.clone(), inflight.delay_until, now);
```
markstory, line 46: Having to clone here could be expensive for larger activation payloads. I don't think that has to block sandbox testing though.

src/store/inflight_activation.rs
```diff
@@ -166,101 +224,196 @@ struct TableRow {
...
+        with_data: bool,
+    ) -> Result<Option<InflightActivation>, Error> {
+        let result = self.get_meta_by_id(id).await?;
+        if result.is_none() {
+            return Ok(result);
+        }
+
+        let mut inflight = result.unwrap();
+
+        if self.is_new_read() && with_data && self.is_new_read() {
```
markstory, line 407:
```suggestion
        if self.is_new_read() && with_data {
```

src/store/inflight_activation.rs
```diff
@@ -463,8 +698,14 @@ impl InflightActivationStore {
...
         sqlx::query("DELETE FROM inflight_taskactivations WHERE id = $1")
             .bind(id)
-            .execute(&self.write_pool)
+            .execute(&self.meta_write_pool)
             .await?;
+
+        sqlx::query("DELETE FROM taskactivation_blobs WHERE id = $1")
+            .bind(id)
+            .execute(&self.blob_write_pool)
+            .await?;
```
markstory, line 707: Does this need an `if self.is_new_write()` guard?

evanh: It doesn't, the table will exist regardless of the guards. The guards just determine what is in the table.

src/store/inflight_activation.rs
```diff
@@ -290,11 +444,85 @@ impl InflightActivationStore {
...
+                .into_iter()
+                .map(BlobTableRow::try_from)
+                .collect::<Result<Vec<BlobTableRow>, _>>()?;
+            let blob_query = blob_query_builder
+                .push_values(blob_rows, |mut b, row| {
+                    b.push_bind(row.id);
+                    b.push_bind(row.activation);
+                    b.push_bind(row.added_at.timestamp());
+                })
+                .push(" ON CONFLICT(id) DO NOTHING")
```
enochtangg, line 511: Just a note on `ON CONFLICT(id) DO NOTHING` since I think it has an interesting implication down the road. We know that it's possible for writes to the blob db to succeed and writes to meta db to fail. If a batch fails and is retried later, the blob will not be updated since the ID already exists. This seems ok since the blob doesn't store any information that gets altered by `inflight_activation_writer` when it gets retried.

enochtangg (commented):
Nice stuff! This is quite a fundamental change so we should do thorough testing in the sandbox before we consider pushing to production

PR #372, opened 2025-05-17
feat(consumer): Batch writes base on payload size instead of elements

john-z-yang:


src/config.rs
```diff
@@ -177,7 +181,8 @@ impl Default for Config {
             kafka_send_timeout_ms: 500,
             db_path: "./taskbroker-inflight.sqlite".to_owned(),
             db_write_failure_backoff_ms: 4000,
-            db_insert_batch_size: 1024,
+            db_insert_batch_size_rows: 256,
```
markstory, line 184: This change in defaults will impact existing deployments. We'd need to update the configuration of existing deployments before this could be merged.

src/kafka/inflight_activation_batcher.rs
```diff
@@ -73,18 +78,25 @@ impl Reducer for InflightActivationBatcher {
...
         }
+
+        metrics::histogram!("consumer.batch_size_rows").record(self.buffer.len() as f64);
+        metrics::histogram!("consumer.batch_size_bytes").record(self.buffer_size as f64);
+
+        self.buffer_size = 0;
         Ok(Some(replace(
             &mut self.buffer,
-            Vec::with_capacity(self.config.max_buf_len),
+            Vec::with_capacity(self.config.max_buf_size),
```
markstory, line 88: Should this be `max_buf_rows` as this was a row based length before.

src/kafka/inflight_activation_writer.rs
```diff
@@ -30,7 +30,7 @@ impl ActivationWriterConfig {
     /// Convert from application configuration into InflightActivationWriter config.
     pub fn from_config(config: &Config) -> Self {
         Self {
-            max_buf_len: config.db_insert_batch_size,
+            max_buf_len: config.db_insert_batch_size_bytes,
```
markstory, line 33: Should this be `config. db_insert_batch_size_rows` as this was a row count before.

src/kafka/inflight_activation_batcher.rs
john-z-yang: Yes, fixed, thanks

src/kafka/inflight_activation_writer.rs
john-z-yang: Good catch

markstory (approved):
Looks good to me.

evanh (approved):
Looks good, nicely done!

john-z-yang:
@cmanallen 

> How does this bound disk usage?

This bounds the disk usage of each insert query, not the total disk usage.

> Capping the insert size may help read queries interleave better between writes

It doesn't. Sqlite with WAL let us have read queries even when write lock is held.

This change helps us interleave other **_write_** queries. When the batch size is very large (i.e. individual messages becomes larger), other write queries cannot run for a very long time since as of now we always naively inserts N rows at a time. This is because we force an fsync like you mentioned to meet our durability requirement.

> Can we increase the autocheckpoint size and checkpoint less frequently? That would means tasks are delayed by some amount if the queue is empty but that our writes are less disruptive during high intensity read-write periods. You'll still fsync the WAL with "PRAGMA synchronous FULL" for full durability.

Sorry, I am not sure what problem this is solving?

cmanallen:
@john-z-yang 

> Sorry, I am not entirely sure what problem this is solving?

Not one you're solving!  The problem description didn't have the full details for outside observers so I had to intuit one.  Now reading "resulting in workers unable to retrieve activations" my understanding is we want to decrease insert latency not improve write throughput or read/write concurrency.  Which makes more sense why you chose the solution you did.

PR #375, opened 2025-05-20
feat(sqlite): Remove the need for the protobuf object in processing

evanh:
Move any fields that are fetched from the protobuf object and move them up to the table, so none of
the processing needs the actual protobuf to run. This ensures that encoding/decoding is only done at
the edges of the taskbroker.


src/store/inflight_activation.rs
```diff
@@ -166,66 +192,111 @@ struct TableRow {
...
             processing_attempts: value.processing_attempts,
             expires_at: value.expires_at,
             delay_until: value.delay_until,
-            processing_deadline_duration: value.activation.processing_deadline_duration as u32,
+            processing_deadline_duration: value.processing_deadline_duration,
             processing_deadline: value.processing_deadline,
             status: value.status,
             at_most_once: value.at_most_once,
             namespace: value.namespace.clone(),
+            taskname: value.taskname.clone(),
```
john-z-yang, line 218: You might be able to get away with moving instead of cloning since `try_from` gets full ownership of `value: InflightActivation`
```rust
namespace: value.namespace.clone(),
taskname: value.taskname.clone(),
```

migrations/0003_update_activations_table.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE inflight_taskactivations ADD on_attempts_exceeded TEXT DEFAULT 'Discard';
+ALTER TABLE inflight_taskactivations ADD received_at INTEGER NOT NULL;
+ALTER TABLE inflight_taskactivations ADD taskname TEXT NOT NULL;
```
john-z-yang, line 3: NIT: need a new line here

src/store/inflight_activation.rs
```diff
@@ -74,10 +56,48 @@ impl From<TaskActivationStatus> for InflightActivationStatus {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Type)]
+pub enum InflightOnAttemptsExceeded {
+    /// Unused but necessary to align with sentry-protos
+    Unspecified,
+    Discard,
+    Deadletter,
+}
```
john-z-yang, line 65: I think I probably missed something, but how come we cannot use the enum from `sentry-protos`? Is it due us not able to implement FromRow because of the orphan rule?

src/grpc/server.rs
```diff
@@ -36,14 +37,14 @@ impl ConsumerService for TaskbrokerServer {
...
-                        "taskname" => inflight.activation.taskname.clone(),
+                        "namespace" => inflight.namespace,
+                        "taskname" => inflight.taskname,
                     )
                     .record(received_to_gettask_latency as f64);
                 }
 
                 let resp = GetTaskResponse {
-                    task: Some(inflight.activation),
+                    task: Some(TaskActivation::decode(&inflight.activation as &[u8]).unwrap()),
```
john-z-yang, line 47: I wonder if we can just give out this vec<u8> here instead of decoding it. Because the server middleware likely just serialize it again.

src/store/inflight_activation.rs
```diff
@@ -148,16 +173,17 @@ impl From<SqliteQueryResult> for QueryResult {
 
 pub struct FailedTasksForwarder {
     pub to_discard: Vec<String>,
-    pub to_deadletter: Vec<TaskActivation>,
+    pub to_deadletter: Vec<String>,
 }
 
 #[derive(Debug, FromRow)]
-struct TableRow {
+struct MetaTableRow {
```
john-z-yang, line 180: I think it might be worthwhile to keep this as `TableRow` for readability reason as we still only have a single table right now

john-z-yang (commented):
Looking good so far, I have some questions and minor nits. Good work!

src/grpc/server.rs
evanh: This is something I need to look into but I don't think our gRPC library will allow it. It's frustrating that I have to decode something that will immediately be encoded again though.

markstory: While it's not ideal this isn't new behavior, we've always been deserializing and reserializing :cry:

src/grpc/server.rs
```diff
@@ -97,11 +98,10 @@ impl ConsumerService for TaskbrokerServer {
 
         if let Ok(Some(inflight_activation)) = update_result {
             let duration = Utc::now() - inflight_activation.added_at;
-            let activation = inflight_activation.activation;
             metrics::histogram!(
                 "task_execution.conclusion.duration",
-                "namespace" => activation.namespace.clone(),
-                "taskname" => activation.taskname.clone(),
+                "namespace" => inflight_activation.namespace.clone(),
+                "taskname" => inflight_activation.taskname.clone(),
```
markstory, line 104: Do we still need to clone here?

src/kafka/inflight_activation_writer.rs
```diff
@@ -197,61 +193,78 @@ mod tests {
...
+                .encode_to_vec(),
                 status: InflightActivationStatus::Pending,
                 partition: 0,
                 offset: 0,
                 added_at: Utc::now(),
+                received_at: DateTime::from_timestamp(
+                    received_at.seconds,
+                    received_at.nanos as u32,
+                )
+                .expect(""),
```
markstory, line 224:
```suggestion
                .unwrap(),
```

This could be unwrap() as you're not providing a message to expect().

src/store/inflight_activation.rs
markstory: > Is it due us not able to implement FromRow because of the orphan rule?

Yes. :cry:

src/store/inflight_activation.rs
```diff
@@ -166,66 +192,111 @@ struct TableRow {
...
-            .max_connections(64)
-            .connect_with(
-                SqliteConnectOptions::from_str(url)?
-                    .journal_mode(SqliteJournalMode::Wal)
-                    .synchronous(SqliteSynchronous::Normal)
-                    .read_only(true)
-                    .disable_statement_logging(),
-            )
-            .await?;
+        sqlx::migrate!("./migrations").run(&write_pool).await?;
```
markstory, line 299: We also run migrations on line 312 do we need to do it twice?

src/store/inflight_activation.rs
```diff
@@ -659,7 +742,7 @@ impl InflightActivationStore {
             {
                 forwarder.to_discard.push(activation.id)
             } else if retry_state.on_attempts_exceeded == OnAttemptsExceeded::Deadletter as i32 {
-                forwarder.to_deadletter.push(activation)
+                forwarder.to_deadletter.push(activation.id)
```
markstory, line 745: Would we be able to revise this logic to not need to deserialize the activation in a future change?

src/store/inflight_activation_tests.rs
```diff
@@ -423,13 +424,16 @@ async fn test_handle_processing_at_most_once() {
...
 
     batch[1].status = InflightActivationStatus::Processing;
-    batch[1].activation.retry_state = Some(RetryState {
-        attempts: 0,
-        max_attempts: 1,
-        on_attempts_exceeded: OnAttemptsExceeded::Discard as i32,
-        at_most_once: Some(true),
-        delay_on_retry: None,
-    });
+    replace_retry_state(
```
markstory, line 427: Nice :clap:

src/upkeep.rs
```diff
@@ -189,27 +190,41 @@ pub async fn do_upkeep(
...
+                        }
+                    };
+
+                    metrics::histogram!("upkeep.dlq.message_size")
+                        .record(inflight.activation.len() as f64);
                     let delivery = producer
                         .send(
                             FutureRecord::<(), Vec<u8>>::to(&config.kafka_deadletter_topic)
-                                .payload(&payload),
+                                .payload(&inflight.activation),
```
markstory, line 216: Nice that we get one fewer deserialize/serialize cycle here.

src/store/inflight_activation.rs
evanh: I did this as a precursor for when we have a separate blob storage.

src/store/inflight_activation.rs
evanh: We can't implement the From<u32> for the taskbroker enum because we don't own that enum. So we need our type instead. The same thing for `InflightActivationStatus`.

src/grpc/server.rs
evanh: Yes, those values get used again further down.

src/store/inflight_activation.rs
evanh: Oops bad merge.

src/store/inflight_activation.rs
evanh: Oops yeah I had that but it got merged out. Fixed.

src/store/inflight_activation.rs
john-z-yang: Since `OnAttemptsExceeded` is `#[repr(i32)]`. How about something like this?

Please let me know because I did not try this.

```rust
#[derive(Debug, FromRow)]
struct TableRow {
    ...
    #[sqlx(try_from = "i32")]
    on_attempts_exceeded: OnAttemptsExceeded,
}
```

src/store/inflight_activation.rs
john-z-yang: It's better to rename it to `MetaTableRow` in the pr where you actually do the separation, not here then?

evanh: Sure, if you like. Just causes more work in the future.

src/store/inflight_activation.rs
evanh: That totally worked. Too bad we didn't know about this earlier, we could have used it for the status of the tasks too. Since that column is already a string, I want to try if `try_from = "String"` would work. 🤔

src/upkeep.rs
```diff
@@ -112,8 +111,10 @@ pub async fn do_upkeep(
             .map(|inflight| {
                 let producer = producer.clone();
                 let config = config.clone();
+                // let store = store.clone();
```
enochtangg, line 114:
```suggestion
```

markstory (approved):
Looks good to me.

PR #379, opened 2025-05-22
Revert "Revert "feat(sqlite): Remove the need for the protobuf object…

evanh:
Fix the migrations and add a test for further migrations.

### DIFF
--- a/src/store/inflight_activation.rs
+++ b/src/store/inflight_activation.rs
@@ -209,7 +209,7 @@ impl From<TableRow> for InflightActivation {
     }
 }
 
-async fn create_sqlite_pool(url: &str) -> Result<(Pool<Sqlite>, Pool<Sqlite>), Error> {
+pub async fn create_sqlite_pool(url: &str) -> Result<(Pool<Sqlite>, Pool<Sqlite>), Error> {
     if !Sqlite::database_exists(url).await? {
         Sqlite::create_database(url).await?
     }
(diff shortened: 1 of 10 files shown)


src/store/inflight_activation_tests.rs
```diff
@@ -783,3 +808,137 @@ async fn test_clear() {
...
+        Ok(TestFolders {
+            parent_folder,
+            initial_folder,
+            other_folder,
+        })
+    }
+}
+
+impl Drop for TestFolders {
+    fn drop(&mut self) {
```
markstory, line 847: This is a neat way to do test teardown :clap:

markstory (approved):
Looks good to me. Nice work on the migrations tests. :clap: They should help prevent this class of problem in the future too.

src/store/inflight_activation.rs
```diff
@@ -166,77 +156,110 @@ struct TableRow {
     status: InflightActivationStatus,
     at_most_once: bool,
     namespace: String,
+    taskname: String,
+    #[sqlx(try_from = "i32")]
+    on_attempts_exceeded: OnAttemptsExceeded,
```
john-z-yang, line 161: I took another look at how we're using `InflightActivationStatus`, and I realize there's a state that we do not have in `ActivationStatus`, which is `delay`.

This makes sense because `ActivationStatus` is for communication between worker and broker over the wire. While `InflightActivationStatus` is for book keeping within the broker itself.

So using `OnAttemptsExceeded` here means that in the future we're bounding the possible types to the wire format. Which has its implications.

I don't think I know enough about the guts of our grpc service to say one is correct over the other.
So, I leave it entirely in your hands.

john-z-yang: Sounds good

PR #396, opened 2025-06-03
feat: Enable deploys for US region for control silo

markstory:
The control silo brokers will live in the US region, so we'll need deploys for US region enabled.


evanh (approved):

evanh (approved):
😨

Issue #267, opened 2025-03-19
Enable taskbroker and taskworker in self-hosted


markstory:
Before we can remove any shim logic from any tasks, we'll need to ensure that taskbroker and a taskworker are included in self-hosted docker-compose configuration.

We should also update the documentation for self-hosted to document that additional containers will be required going forward.

aldy505:
Sound this on Discord https://discord.com/channels/621778831602221064/796028405833007104/1379660189656944700

PR #398, opened 2025-06-04
feat: Add release workflow for creating dockerhub images

markstory:
- Add release workflow based on how vroom does it.
- Update dockerfile to remove unused arg and set default for version

Refs #267

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ COPY ./src ./src
 RUN rm ./target/release/deps/taskbroker*
 RUN cargo build --release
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -32,7 +30,7 @@ COPY ./src ./src
 RUN rm ./target/release/deps/taskbroker*
 RUN cargo build --release
 
-RUN echo "${TASKBROKER_VERSION}" > ./VERSION
+RUN echo "${TASKBROKER_VERSION:-nightly}" > ./VERSION
```
evanh, line 33: Why does this need `-nightly`?

evanh (approved):

PR #409, opened 2025-06-09
Fix image workflow

markstory:


aldy505 (commented):
There is actually a better way, but I never seen this being used on sentry org: https://github.com/docker/login-action

markstory:
> There is actually a better way, but I never seen this being used on sentry org

Me neither, I'd rather not be the first, as our build tooling has enough ❄️ right now.

evanh (approved):

PR #412, opened 2025-06-12
ci: build arm64 image (second attempt)

aldy505:
Hopefully this would work a lot better

### DIFF
--- a/.github/workflows/image.yml
+++ b/.github/workflows/image.yml
@@ -105,11 +105,7 @@ jobs:
 
       - shell: bash
         run: |
-          IMAGE_URL="ghcr.io/getsentry/taskbroker:${{ github.sha }}"
           echo "${{ secrets.DOCKER_HUB_RW_TOKEN }}" | docker login --username=sentrybuilder --password-stdin
-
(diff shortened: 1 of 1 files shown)


.github/workflows/image.yml
```diff
@@ -56,15 +113,14 @@ jobs:
 
           # We push 3 tags to Dockerhub:
           # first, the full sha of the commit
-          docker tag "$IMAGE_URL" getsentry/taskbroker:${GITHUB_SHA}
-          docker push getsentry/taskbroker:${GITHUB_SHA}
+          docker buildx imagetools create --tag getsentry/taskbroker:${GITHUB_SHA} ghcr.io/getsentry/taskbroker:${{ github.sha }}
```
markstory, line 116: If we've fetched the image from ghcr, why do we need to use `imagetools create` on it?

aldy505: Ah right. The docker pull should be removed first. See https://github.com/getsentry/symbolicator/pull/1716#discussion_r2143076552

markstory (approved):

PR #413, opened 2025-06-12
ci: lightweight dockerfile using bookworm-slim

aldy505:
The runtime image shouldn't use `rust` as the image base, since we don't need any `cargo` or `rustc` binary. Instead, we can just use `bookworm-slim` and install sqlite3 dev headers there.


Dockerfile.dev
```diff
@@ -0,0 +1,55 @@
+# Build image
```
evanh, line 1: What is this file for?

evanh (commented):
/gcbrun

Dockerfile.dev
aldy505: If you don't have any importance/relying on `rust` being the base image, we can delete this

aldy505: Oh, so that you can still use it if you depends on `rust` being base image.

markstory: Where would it be used? The images we use in self-hosted, and saas all come from the standard `Dockerfile`. Could we remove this dockerfile as it won't be used?

aldy505:
@markstory done!

markstory:
This had to be reverted as the runtime image had broken SSL

PR #424, opened 2025-06-19
fix: Improve assertions on status counts

markstory:
Add more exhaustive assertions for status counts. This could be helpful in the future if we need to refactor InflightActivationStore is implemented.


evanh (approved):
This makes sense.

PR #428, opened 2025-06-23
feat(logging): Send tracing events at or above INFO to Sentry as logs instead of breadcrumbs

lcian:
- Bump the SDK
- Send everything above INFO as logs instead of breadcrumbs. Keep sending errors as events as per default behavior
- Explicitly list the feature flags to enable: all the default ones + `tracing` + `logs`. Adding `release-health` would add unnecessary dependencies to the lockfile for a feature we're not using. See https://github.com/getsentry/sentry-rust/issues/805#issuecomment-2879047123 and https://github.com/rust-lang/cargo/issues/10801 for more info

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1577,7 +1577,7 @@ dependencies = [
  "num-integer",
  "num-iter",
  "num-traits",
- "rand",
+ "rand 0.8.5",
  "smallvec",
  "zeroize",
 ]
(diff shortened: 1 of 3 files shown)


lcian:
Previously I forgot to enable `enable_logs` in the SDK `init` call, now I've added it.
@markstory Let me know if you prefer me to add it to the `LoggingConfig` instead.

markstory:
> Let me know if you prefer me to add it to the LoggingConfig instead.

I don't think we need that right now. Taskbroker logs are low volume, and forwarding them to sentry shouldn't incur much overhead.

Issue #440, opened 2025-07-09
Improve runbooks for out of space failure mode


markstory:
Update our runbooks to cover how to handle and recover from an 'out of space error' with persistent volumes

PR #442, opened 2025-07-11
feat: Add ability to perform a full vacuum on startup

markstory:
Should a broker run out of storage space, it is difficult for us to force a manual vacuum as if the container is running the application will have a write lock on the database. With these changes we can enable a configuration via k8s that will force the broker to do a vacuum during startup. This should help mitigate out of storage space issues in the future.

Refs #440

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -78,9 +78,9 @@ async fn main() -> Result<(), Error> {
         .await?;
     }
     if config.full_vacuum_on_start {
-        println!("Running full vacuum on database");
+        info!("Running full vacuum on database");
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -77,6 +77,13 @@ async fn main() -> Result<(), Error> {
         )
         .await?;
     }
+    if config.full_vacuum_on_start {
+        println!("Running full vacuum on database");
```
evanh, line 81: Can we log instead of println?

evanh (approved):
Is the plan to have this enabled by default? I can't think of a reason not to have this.

PR #450, opened 2025-07-22
ci: Fix fast-revert bot name

evanpurkhiser:
Without the `[bot]` suffix it will lookup `1330327` as a user id, which
is why were were getting @MineCraftSpy as the reverting user


Jeffreyhung (approved):
LGTM!

PR #452, opened 2025-07-25
feat: Add processing deadline grace period

markstory:
Because of clock skew, and overloaded brokers we can end up resetting deadlines in the brokers while workers don't terminate tasks. This leads to duplicate execution and reduces our ability to make progress.

By giving workers an additional 3 seconds of runtime we should be able to work around most clock skew, and many slow broker scenarios. I chose 3 seconds arbitrarily, and am open to changing it.

I've enabled this behavior by default as I think it will be relatively safe as it reduces strictness.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -164,7 +164,7 @@ pub struct Config {
     /// The number of additional seconds that processing deadlines
     /// are extended by. This helps reduce broker deadline resets when
     /// brokers are under load, or there are small networking delays.
-    pub processing_deadline_grace_sec: i64,
+    pub processing_deadline_grace_sec: u64,
 
     /// The frequency at which upkeep tasks
     /// (discarding, retrying activations, etc.) are executed.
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -161,6 +161,11 @@ pub struct Config {
     /// the activation will be discarded/deadlettered.
     pub max_processing_attempts: usize,
 
+    /// The number of additional seconds that processing deadlines
+    /// are extended by. This helps reduce broker deadline resets when
+    /// brokers are under load, or there are small networking delays.
+    pub processing_deadline_grace_sec: i64,
```
john-z-yang, line 167: How come this is signed instead of something like u64? Do we see a scenario where we want to configure the broker with a negative grace period?

markstory: We don't have a scenario where we would want a negative number. I'll update to u64.

PR #461, opened 2025-08-05
feat: Add a deadline quiet period after startup

markstory:
When taskbrokers are restarted, workers are unable to publish results, and the results they do have get buffered in the worker until the broker is back online. During this restart period, tasks with short deadlines will expire and be made pending again after a broker's first upkeep is completed.

By having an upkeep 'quiet period' we give workers time to flush results which should reduce the number of broker-side deadline resets without a worker kill that we see. This could create a heavier first upkeep if many deadlines have been exceeded, but I think the benefits are worth the tradeoff.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -165,6 +165,11 @@ pub struct Config {
     /// (discarding, retrying activations, etc.) are executed.
     pub upkeep_task_interval_ms: u64,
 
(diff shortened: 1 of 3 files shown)


enochtangg (approved):
LGMT! Lets try it out

PR #466, opened 2025-08-12
feat(taskworker): Purge killswitched tasks in sqlite

enochtangg:
When a task is added to the killswitch runtime config, we also want to purge the task from sqlite. During inc-1308, we used killswitches to drop a poor performing task in kafka , but the offending task persisted in sqlite which delayed recovery.

Refs: https://linear.app/getsentry/issue/STREAM-352/killswitching-a-task-should-also-purge-sqlite

### DIFF
--- a/src/kafka/consumer.rs
+++ b/src/kafka/consumer.rs
@@ -741,7 +741,7 @@ pub async fn reduce_err(
     Ok(())
 }
 
-trait CommitClient {
+pub trait CommitClient {
     fn store_offsets(&self, tpl: &TopicPartitionList) -> KafkaResult<()>;
 }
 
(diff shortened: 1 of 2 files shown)


evanh (approved):
LGTM. Nice!

src/kafka/consumer.rs
```diff
@@ -741,7 +741,7 @@ pub async fn reduce_err(
     Ok(())
 }
 
-trait CommitClient {
+pub trait CommitClient {
```
john-z-yang, line 744: I don't think this trait is being used anywhere else in this pr, how come it is `pub`?

john-z-yang: Ah wait, never mind, I see

PR #473, opened 2025-08-26
feat(ci): Phase 2: Cutover to gha for prod images

hubertdeng123:
Depends on https://github.com/getsentry/taskbroker/pull/471

This cuts over GoCD to using images published from gha. It keeps cloudbuild published images around just in case things go wrong and revert needs to happen


markstory (approved):
Looks good to me.

PR #480, opened 2025-09-05
fix(taskbroker): Add the ability to killswitch the health check

evanh:
In the case where the health check is erroneously restarting pods, the health check can be disabled using this killswitch.

### DIFF
--- a/src/upkeep.rs
+++ b/src/upkeep.rs
@@ -380,6 +380,10 @@ pub async fn check_health(
 ) -> Instant {
     let now = Instant::now();
     if config.health_check_killswitched {
+        metrics::counter!("upkeep.health", "status" => "killswitched").increment(1);
+        health_reporter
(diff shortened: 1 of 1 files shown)


src/upkeep.rs
```diff
@@ -379,6 +379,9 @@ pub async fn check_health(
     mut health_reporter: HealthReporter,
 ) -> Instant {
     let now = Instant::now();
+    if config.health_check_killswitched {
+        return now;
```
fpacifici, line 383: Should you set the status to "serving" in this case ?
```
health_reporter
            .set_service_status(SERVICE_NAME, ServingStatus::Serving)
            .await;
```

So that, if the status is "not serving" and somebody triggers the killswitch k8s would not interpret the pod as being in not serving state.

fpacifici (approved):

PR #485, opened 2025-09-15
feat(sqlite): Add sqlite db status metrics

enochtangg:
Emit status metrics via sqlite3_db_status to track cache usage, hit/miss counters, lookaside allocator, schema/statement memory, and cache write/spill totals.

### DIFF
--- a/src/store/inflight_activation.rs
+++ b/src/store/inflight_activation.rs
@@ -322,141 +322,141 @@ impl InflightActivationStore {
             return;
         }
 
-        if let Ok(mut conn) = self.read_pool.acquire().await {
-            if let Ok(mut raw) = conn.lock_handle().await {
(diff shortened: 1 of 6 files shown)


evanh (approved):

src/store/inflight_activation.rs
```diff
@@ -303,9 +313,155 @@ impl InflightActivationStore {
...
         Ok(())
     }
 
+    async fn emit_db_status_metrics(&self) {
+        if !self.config.enable_sqlite_status_metrics {
+            return;
+        }
+
+        if let Ok(mut conn) = self.read_pool.acquire().await
+            && let Ok(mut raw) = conn.lock_handle().await
```
fpacifici, line 326: Does it means, producing metrics takes the DB lock? Do we need that ?

PR #492, opened 2025-09-24
chore(local dev): Allow for custom GRPC port to be specified

NicoHinderling:
I and some other coworkers had an issue where colima was using port 50051, which is the port that taskbroker also wants to use. Changing the port that taskbroker uses caused tasks to start flowing again

This is an attempt to use port 50055 solely in the devservices/local development environment


devservices/config.yml
```diff
@@ -25,8 +25,9 @@ services:
     environment:
       TASKBROKER_KAFKA_CLUSTER: "kafka-kafka-1:9093"
       TASKBROKER_CREATE_MISSING_TOPICS: "true"
+      TASKBROKER_GRPC_PORT: "50055"
```
NicoHinderling, line 28: the only problem is getting the port to be set to 50055 for local development when running taskbroker standalone. Putting an `EXPORT TASKBROKER_GRPC_PORT=50055` in either `.envrc` or a new `.env` file did not work

I think the current approach should work for the docker container served from devservices, but I think this needs to still work for local taskbroker running side by side without requiring manual intervention from the dev

not sure the best way to accomplish that

Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN groupadd taskbroker --gid 1000 && useradd --gid taskbroker --uid 1000 taskbr
 RUN apt-get update && apt-get upgrade -y && apt-get install -y libssl-dev libsqlite3-dev libsqlite3-0 openssl ca-certificates
 
 EXPOSE 50051
+EXPOSE 50055
```
NicoHinderling, line 44: expose both since IIRC the docker image is used for both cloudbuild (prod) and GHCR (devservices aka local)

PR #495, opened 2025-10-08
ci: tag latest version only for calver release

aldy505:
`latest` should point at CalVer release, and `nightly` should point at master/main branch. We're cleaning things up for self-hosted users.


hubertdeng123:
closing as my follow up was merged

PR #501, opened 2025-10-16
feat(blq): Add runtime option to forward tasks to long namespace

lvthanh03:
Introduces a runtime option to forward tasks in demoted namespaces (defined in runtime config) to the `long` namespace.

Refs STREAM-279

### DIFF
--- a/src/kafka/inflight_activation_batcher.rs
+++ b/src/kafka/inflight_activation_batcher.rs
@@ -14,7 +14,6 @@ use super::consumer::{
     ReduceConfig, ReduceShutdownBehaviour, ReduceShutdownCondition, Reducer,
     ReducerWhenFullBehaviour,
 };
-use tracing::error;
 
 pub struct ActivationBatcherConfig {
     pub kafka_config: ClientConfig,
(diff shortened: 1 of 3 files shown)


evanh (changes requested):
This is not a sufficient change. The task can't just be updated in the local DB, it has to be produced to the topic that the "long" namespace is consuming from. The purpose of this change is to make a different broker pool burn down the backlog. This change will still have the original brokers process the message, the messages will just be marked with a different namespace.

lvthanh03:
> This is not a sufficient change. The task can't just be updated in the local DB, it has to be produced to the topic that the "long" namespace is consuming from. The purpose of this change is to make a different broker pool burn down the backlog. This change will still have the original brokers process the message, the messages will just be marked with a different namespace.

That is correct, I'll look into how to implement this!

src/kafka/inflight_activation_batcher.rs
```diff
@@ -70,6 +91,33 @@ impl Reducer for InflightActivationBatcher {
...
+                    FutureRecord::<(), Vec<u8>>::to(&self.config.kafka_long_topic)
+                        .payload(&t.activation),
+                    Timeout::After(Duration::from_millis(self.config.send_timeout_ms)),
+                )
+                .await;
+            match delivery {
+                Ok(_) => return Ok(()),
+                Err((err, _msg)) => {
+                    error!("inflight_activation_batcher.publish.failure: {}", err);
+                    self.batch_size += t.activation.len();
```
evanh: I would remove these three lines (114 - 116) and just let the function fall through to the normal return on lines 121-124.

src/kafka/inflight_activation_batcher.rs
```diff
@@ -70,6 +91,33 @@ impl Reducer for InflightActivationBatcher {
             return Ok(());
         }
 
+        if runtime_config.demoted_namespaces.contains(namespace) {
```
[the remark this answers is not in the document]
lvthanh03: done, added `runtime_config.demoted_topic`, if left empty/not present in yaml then `config.kafka_long_topic` will be used

src/kafka/inflight_activation_batcher.rs
lvthanh03: This should be fixed as well, I removed the Err clause so it would just fall through below if the produce did not go through.

PR #504, opened 2025-10-24
fix(blq): Fix forward loop in demoted namespaces

lvthanh03:
Checks whether topics that taskbroker is consuming from and producing to is the same before forwarding tasks.

Added metrics for forwarded tasks and fixed inflated completed tasks metrics.

Also handles edge case where all tasks are forwarded and batch is empty.

Tested and working in sandbox.

Refs STREAM-279

### DIFF
--- a/src/kafka/inflight_activation_batcher.rs
+++ b/src/kafka/inflight_activation_batcher.rs
@@ -130,15 +130,12 @@ impl Reducer for InflightActivationBatcher {
             return Ok(None);
         }
 
-        let runtime_config = self.runtime_config_manager.read().await;
-
-        if !self.batch.is_empty() {
-            metrics::histogram!("consumer.batch_rows").record(self.batch.len() as f64);
-            metrics::histogram!("consumer.batch_bytes").record(self.batch_size as f64);
-        }
+        metrics::histogram!("consumer.batch_rows").record(self.batch.len() as f64);
+        metrics::histogram!("consumer.batch_bytes").record(self.batch_size as f64);
 
         // Send all forward batch in parallel
         if !self.forward_batch.is_empty() {
+            let runtime_config = self.runtime_config_manager.read().await;
             let forward_topic = runtime_config
                 .demoted_topic
                 .clone()


src/upkeep.rs
```diff
@@ -300,22 +301,28 @@ pub async fn do_upkeep(
                     let topic = runtime_config.demoted_topic.clone().unwrap_or(config.kafka_long_topic.clone());
 
                     async move {
-                        if inflight.status == InflightActivationStatus::Complete {
-                            return Ok(inflight.id);
-                        }
                         metrics::counter!("upkeep.forward_demoted_namespace", "namespace" => inflight.namespace, "taskname" => inflight.taskname).increment(1);
+
+                        let modified_activation = match modify_activation_namespace(&inflight.activation) {
```
fpacifici: This does not mutate `inflight` in place right? It creates a new activation with the changed namespace instead. Is it correct ?  (as `inflight` is used later in this method, mutating it in place would be dangerous)

src/kafka/inflight_activation_batcher.rs
```diff
@@ -91,32 +95,24 @@ impl Reducer for InflightActivationBatcher {
...
+                    )
+                    .increment(1);
+                    self.forward_batch.push(tagged_activation);
+                    return Ok(());
+                }
+                Ok(None) => {
+                    // Already forwarded, fall through to add to batch
+                }
+                Err(_) => {
+                    // Decode error, fall through to add to batch to handle in upkeep
```
evanh: We should have a metric at least for this case, so we can tell if something weird is happening.

src/kafka/inflight_activation_batcher.rs
```diff
@@ -127,12 +123,40 @@ impl Reducer for InflightActivationBatcher {
...
         }
 
-        metrics::histogram!("consumer.batch_rows").record(self.batch.len() as f64);
-        metrics::histogram!("consumer.batch_bytes").record(self.batch_size as f64);
+        let runtime_config = self.runtime_config_manager.read().await;
+
+        if !self.batch.is_empty() {
+            metrics::histogram!("consumer.batch_rows").record(self.batch.len() as f64);
+            metrics::histogram!("consumer.batch_bytes").record(self.batch_size as f64);
+        }
```
markstory: Why not always make the metrics? The zeros are meaningful.

src/kafka/inflight_activation_batcher.rs
```diff
@@ -127,12 +123,40 @@ impl Reducer for InflightActivationBatcher {
...
+            let topic = runtime_config
+                .demoted_topic
+                .clone()
+                .unwrap_or(self.config.kafka_long_topic.clone());
+            let sends = self.forward_batch.iter().map(|payload| {
+                self.producer.send(
+                    FutureRecord::<(), Vec<u8>>::to(&topic).payload(payload),
+                    Timeout::After(Duration::from_millis(self.config.send_timeout_ms)),
+                )
+            });
```
markstory: One concern I have with demoted namespaces, is that the producer cluster won't always have a long topic deployed on it. For example, if a broker is connected to `kafka-events` and we need to demote to the long/limited topic, that topic only exists on `kafka-small`.  The current logic will result in the broker failing to produce.

I think we would need to expand the application configuration to have broker configuration for the demoted namespace flow.  That would also imply that the topic we demote to is also fixed and could be removed from runtime config.

src/kafka/inflight_activation_batcher.rs
```diff
@@ -127,12 +123,40 @@ impl Reducer for InflightActivationBatcher {
...
+
+            let results = join_all(sends).await;
```
markstory: Much better :clap:

lvthanh03:
Switched to comparing topic values to determine whether the task needs forwarding. Also added more metrics for visibility.

PR #508, opened 2025-11-04
feat(v2): PoC of a version of taskbroker that uses Postgres

evanh:
Rewrite the taskbroker to use Postgres. This is just a PoC so the code isn't properly cleaned up yet.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@
 # recent enough version of protobuf-compiler
 FROM rust:1-bookworm AS build
 
-RUN apt-get update && apt-get upgrade -y 
+RUN apt-get update && apt-get upgrade -y
 RUN apt-get install -y cmake pkg-config libssl-dev librdkafka-dev protobuf-compiler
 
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -121,6 +121,12 @@ pub struct Config {
     /// The number of ms for timeouts when publishing messages to kafka.
     pub kafka_send_timeout_ms: u64,
 
+    /// The url of the postgres database to use for the inflight activation store.
+    pub pg_url: String,
+
+    /// The name of the postgres database to use for the inflight activation store.
+    pub pg_database_name: String,
```
markstory: You should be able to combine these, and include the database name into the url.

evanh: I separated it because technically I interact with 2 databases: default, to create the DB, and then the actual DB to create the table.

markstory: Right, because you can't connect to a database that doesn't exist, and you need to create the db initially.

PR #511, opened 2025-11-07
feat(blq): Add runtime config for tasks forward_cluster

lvthanh03:
Adds a runtime config option to specify which Kafka cluster to forward tasks from demoted namespaces to.

Refs STREAM-560


evanh (approved):
Nicely done, looks good to me!

PR #521, opened 2025-12-25
ref(build): release Docker image to GHCR via Craft

aldy505:
This is an ongoing effort to simplify the release process to not rely on the hacky GitHub Actions script that copies the Docker image after a release was created.

Part of https://github.com/getsentry/self-hosted/issues/4123

REF SELF-78

### DIFF
--- a/.github/workflows/build-client.yml
+++ b/.github/workflows/build-client.yml
@@ -1,0 +1,28 @@
+name: Build Python client
+
+on:
+  push:
+    branches:
+      - release/**
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    name: Build and upload artifacts
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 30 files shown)


scripts/bump-version.sh
```diff
@@ -0,0 +1,17 @@
...
+fi
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+cd $SCRIPT_DIR/..
+
+OLD_VERSION="${1}"
+NEW_VERSION="${2}"
+
+echo "Current version: ${OLD_VERSION}"
+echo "Bumping version: ${NEW_VERSION}"
```
markstory: Does craft read the output of this script?

aldy505: Nope. Craft only need this script to run, so the changelog will be properly generated.

markstory:
Looks good, but there is an image build permission issue.

aldy505:
@markstory when you're back, feel free to merge this :)

PR #529, opened 2026-01-13
feat(v2): Postgres storage adapter

evanh:
This adds a postgres storage adapter for the taskbroker, as well as providing a way to choose between the adapters in the configuration. This adapter will also work with AlloyDB.

In postgres, the keyword `offset` is reserved, so that column is called `kafka_offset` in the PG tables and converted to `offset`.

The tests were updated to run with both the SQLite and Postgres adapter using the rstest crate. The `create_test_store` function was updated to be the standard for all tests, and to allow choosing between a SQLite and Postgres DB.

A `remove_db` function was added to the trait and the existing adapters, since the tests create a unique PG database on every run that should be cleaned up.

The `create_test_store` function was updated to be the standard for all tests, and to allow choosing between an SQLite and Postgres DB.

### DIFF
--- a/src/store/inflight_activation.rs
+++ b/src/store/inflight_activation.rs
@@ -1,10 +1,4 @@
 use anyhow::{Error, anyhow};
-use sqlx::postgres::PgQueryResult;
-use std::fmt::Result as FmtResult;
-use std::fmt::{Display, Formatter};
-use std::{str::FromStr, time::Instant};
-
-use crate::config::Config;
 use async_trait::async_trait;
 use chrono::{DateTime, Utc};
 use derive_builder::Builder;
@@ -21,13 +15,17 @@ use sqlx::{
     ConnectOptions, FromRow, Pool, QueryBuilder, Row, Sqlite, Type,
     migrate::MigrateDatabase,
     pool::{PoolConnection, PoolOptions},
+    postgres::PgQueryResult,
     sqlite::{
         SqliteAutoVacuum, SqliteConnectOptions, SqliteJournalMode, SqlitePool, SqliteQueryResult,
         SqliteRow, SqliteSynchronous,
     },
 };
+use std::fmt::{Display, Formatter, Result as FmtResult};
 use std::{str::FromStr, time::Instant};
 use tracing::{instrument, warn};
+
+use crate::config::Config;
 
 /// The members of this enum should be synced with the members
 /// of InflightActivationStatus in sentry_protos
--- a/src/kafka/inflight_activation_writer.rs
+++ b/src/kafka/inflight_activation_writer.rs
@@ -200,31 +200,16 @@ impl Reducer for InflightActivationWriter {
 
 #[cfg(test)]
 mod tests {
-    use super::{ActivationWriterConfig, InflightActivation, InflightActivationWriter, Reducer};
-    use chrono::{DateTime, Utc};
-    use prost::Message;
-    use prost_types::Timestamp;
+    use chrono::DateTime;
     use rstest::rstest;
-    use std::collections::HashMap;
-
-    use crate::test_utils::create_test_store;
-    use sentry_protos::taskbroker::v1::OnAttemptsExceeded;
-    use sentry_protos::taskbroker::v1::TaskActivation;
 
     use super::{ActivationWriterConfig, InflightActivationWriter, Reducer};
-    use crate::store::inflight_activation::InflightActivationStatus;
-
-    use chrono::DateTime;
-    use std::sync::Arc;
-
-    use crate::store::inflight_activation::InflightActivationBuilder;
-    use crate::store::inflight_activation::{
-        InflightActivationStatus, InflightActivationStore, InflightActivationStoreConfig,
-        SqliteActivationStore,
+    use crate::{
+        store::inflight_activation::{InflightActivationBuilder, InflightActivationStatus},
+        test_utils::{
+            TaskActivationBuilder, create_test_store, generate_unique_namespace, make_activations,
+        },
     };
-    use crate::test_utils::TaskActivationBuilder;
-    use crate::test_utils::generate_unique_namespace;
-    use crate::test_utils::make_activations;
 
     #[tokio::test]
     #[rstest]
--- a/src/store/inflight_activation_tests.rs
+++ b/src/store/inflight_activation_tests.rs
@@ -1,34 +1,22 @@
-use prost::Message;
-use rstest::rstest;
-use sqlx::{QueryBuilder, Sqlite};
-use std::collections::HashSet;
-use std::collections::{HashMap, HashSet};
-use std::fs;
-use std::io::Error;
-use std::path::Path;
-use std::sync::Arc;
-use std::time::Duration;
-
-use crate::config::Config;
-use crate::store::inflight_activation::{
-    InflightActivationBuilder, InflightActivationStatus, InflightActivationStore,
-    InflightActivationStoreConfig, QueryResult, SqliteActivationStore, create_sqlite_pool,
-};
-use crate::test_utils::{StatusCount, TaskActivationBuilder};
-use crate::test_utils::{
-    assert_counts, create_integration_config, create_test_store, generate_temp_filename,
-    generate_unique_namespace, make_activations, make_activations_with_namespace,
-    replace_retry_state,
-};
 use chrono::{DateTime, SubsecRound, TimeZone, Utc};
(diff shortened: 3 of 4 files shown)


src/grpc/server_tests.rs
```diff
@@ -157,6 +178,7 @@ async fn test_set_task_status_success() {
         }),
     };
     let response = service.set_task_status(Request::new(request)).await;
+    println!("response: {:?}", response);
```
fpacifici: please remove

src/store/inflight_activation.rs
```diff
@@ -93,7 +125,7 @@ pub struct InflightActivation {
     /// The duration in seconds that a worker has to complete task execution.
     /// When an activation is moved from pending -> processing a result is expected
     /// in this many seconds.
-    pub processing_deadline_duration: u32,
+    pub processing_deadline_duration: i32,
```
fpacifici: Any chance that we have values in the sqlite DB that would not fit a i32 ?

src/store/inflight_activation.rs
```diff
@@ -360,6 +400,9 @@ pub trait InflightActivationStore: Send + Sync {
 
     /// Remove killswitched tasks
     async fn remove_killswitched(&self, killswitched_tasks: Vec<String>) -> Result<u64, Error>;
+
+    /// Remove the database, used only in tests
+    async fn remove_db(&self) -> Result<(), Error>;
```
fpacifici: WE should create an MCP in front of production taskbroker and we absolutely need to expose this, which will certainly be used only in tests.

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+        )
+        .bind(&config.pg_database_name)
+        .fetch_one(&default_pool)
+        .await?;
+
+        if !row.0 {
+            println!("Creating database {}", &config.pg_database_name);
+            sqlx::query(format!("CREATE DATABASE {}", &config.pg_database_name).as_str())
+                .bind(&config.pg_database_name)
+                .execute(&default_pool)
```
fpacifici: nit: the pool is called read_pool inside the `create_default_postgres_pool` method. While that does not make it a read only pool, you probably want to rename it as you would be using it to create a DB

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+                .await?;
+        }
+        // Close the default pool
+        default_pool.close().await;
+
+        let (read_pool, write_pool) =
+            create_postgres_pool(&config.pg_url, &config.pg_database_name).await?;
+        sqlx::migrate!("./pg_migrations").run(&write_pool).await?;
```
fpacifici: Same as above. Here we have a global store, each taskbroker should not be creating the DB. It was alright with sqlite as it was a local store.

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+        Ok(Self {
+            read_pool,
+            write_pool,
+            config,
+        })
+    }
+}
+
+#[async_trait]
+impl InflightActivationStore for PostgresActivationStore {
```
fpacifici: Am I missing something or we do not have an `InflightActivationStore` trait ?

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+                    on_attempts_exceeded
+                )
+            ",
+        );
+        let rows = batch
+            .into_iter()
+            .map(TableRow::try_from)
+            .collect::<Result<Vec<TableRow>, _>>()?;
+        let query = query_builder
+            .push_values(rows, |mut b, row| {
```
fpacifici: Is this going to write the batch in one statement or is this iterating client side ? 
We would want to insert a batch of data in the same statement, postgers is much more efficient at writing this way:

There are a number of ways to do that, this is an example.
```
INSERT INTO users (field, field, field)
SELECT *
FROM (
    VALUES
        ('a1', ''asdas', 'asdasd'),
        ('a1', ''asdas', 'asdasd'),) 
AS v(field, field, field);
```

src/config.rs
```diff
@@ -121,6 +121,14 @@ pub struct Config {
     /// The number of ms for timeouts when publishing messages to kafka.
     pub kafka_send_timeout_ms: u64,
 
+    pub database_adapter: &'static str,
```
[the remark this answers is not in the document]
evanh: I want to pass this as a environment variable, and it seemed cleaner to pass as a string that way.

src/store/inflight_activation.rs
evanh: No, these are all time durations less than an hour I believe.

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    /// Update tasks that are in processing and have exceeded their processing deadline
+    /// Exceeding a processing deadline does not consume a retry as we don't know
+    /// if a worker took the task and was killed, or failed.
+    #[instrument(skip_all)]
+    async fn handle_processing_deadline(&self) -> Result<u64, Error> {
+        let now = Utc::now();
+        let mut atomic = self.write_pool.begin().await?;
+
+        // Idempotent tasks that fail their processing deadlines go directly to failure
```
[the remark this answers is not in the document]
evanh: Yes it is.

markstory:
> I'd recommend reducing the duplication in the ActivationStore between the two stores. I don't know whether we will remove SQLite from the implementation for good, anyway it will take some time before we get there, having so much copy paste is quite dangerous.

Having the duplication is also useful to allow the implementations to diverge as the SQL dialects of postgres and sqlite are different enough to make code that creates queries challenging to share. For non-query logic having common logic pulled out makes sense.

default_migrations/0001_create_database.sql
```diff
@@ -0,0 +1 @@
+CREATE DATABASE taskbroker;
```
markstory: What's this for? You shouldn't need to create a database for postgres or sqlite.

pg_migrations/0001_create_inflight_activations.sql
```diff
@@ -0,0 +1,20 @@
...
+    added_at TIMESTAMPTZ NOT NULL,
+    received_at TIMESTAMPTZ NOT NULL,
+    processing_attempts INTEGER NOT NULL,
+    expires_at TIMESTAMPTZ,
+    delay_until TIMESTAMPTZ,
+    processing_deadline_duration INTEGER NOT NULL,
+    processing_deadline TIMESTAMPTZ,
+    status TEXT NOT NULL,
+    at_most_once BOOLEAN NOT NULL DEFAULT FALSE,
+    application TEXT NOT NULL DEFAULT '',
```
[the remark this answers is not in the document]
markstory: Yeah, Ideally this is a NOT NULL with no default. Updating the sqlite schema is tedious because sqlite doesn't have a way to modify/remove columns from schema without shenanigans.

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+
+use crate::config::Config;
+
+pub async fn create_postgres_pool(
+    url: &str,
+    database_name: &str,
+) -> Result<(Pool<Postgres>, Pool<Postgres>), Error> {
+    let conn_str = url.to_owned() + "/" + database_name;
+    let read_pool = PgPoolOptions::new()
+        .max_connections(64)
```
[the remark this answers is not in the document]
fpacifici: prod-eng already did quite a few tests on AlloyDB. They should already have an idea on whether to use pgbouncer or not

src/store/inflight_activation.rs
markstory: what could go wrong?

src/config.rs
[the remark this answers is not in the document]
evanh: OK I updated this to match how LogFormat is used.

src/store/postgres_activation_store.rs
```diff
@@ -0,0 +1,732 @@
...
+            return Ok(None);
+        }
+        let result = self
+            .get_pending_activations_from_namespaces(application, namespaces.as_deref(), Some(1))
+            .await?;
+        if result.is_empty() {
+            return Ok(None);
+        }
+        Ok(Some(result[0].clone()))
+    }
```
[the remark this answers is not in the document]
evanh: I think the default implementation is the better way to go. I think I will do that in a separate PR though.

PR #531, opened 2026-01-14
feat(client) Add the first chunk of the taskbroker_client

markstory:
This adds the Application, Namespace and Task objects needed to define and send tasks. I'll add the worker and scheduler separately to keep pull requests reasonably sized.

Refs STREAM-605

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
       - uses: getsentry/action-setup-venv@5a80476d175edf56cb205b08bc58986fa99d1725 # v3.2.0
         with:
           cache-dependency-path: uv.lock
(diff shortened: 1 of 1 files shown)


clients/python/pyproject.toml
```diff
@@ -52,33 +52,6 @@ testpaths = ["tests"]
...
-# begin: missing 3rd party stubs
-[[tool.mypy.overrides]]
-module = [
-    ".conftest",
-    "redis.*",
-    "rediscluster.*",
-    "confluent_kafka.*",
-]
-ignore_missing_imports = true
-# end: missing 3rd party stubs
```
markstory: This isn't required as mypy configuration is shared with integration_tests in the top level pyproject.toml file.

PR #536, opened 2026-01-20
feat(client): Add worker implementation and tests

markstory:
Add worker implementation including TaskWorker, client connection handling, and child worker process management. Includes comprehensive test coverage for worker functionality.

Part of STREAM-605


markstory:
Closing as this can be split up better.

PR #559, opened 2026-02-19
feat: Separate runtimes for taskbroker components

evanh:
Create a separate tokio runtime for each of the consumer and gRPC server. The maintenance and upkeep threads will continue to use the default runtime. The number of threads allocated to each runtime is specified by a config setting, or the TOKIO_WORKER_THREADS environment variable or defaults to the number of CPUs.


evanh:
> I'm interested to see how this impacts contention on sqlite, or if the connection can be shared across multiple runtimes.

I doubt it will have much impact on SQLite, we would still be locked at one process at a time. This change is meant more for the push taskbroker testing, so we can isolate the push threads from the other components.

PR #570, opened 2026-03-10
feat: Add simpler API for creating external tasks

markstory:
While it was possible to create external tasks by creating a standalone registry, and using it to define tasks, these changes make creating external namespaces and tasks more ergonomic and provide conventions for routing tasks between applications.

Refs STREAM-611


markstory:
> What is the scenario where an application would want to submit tasks for another application?

The upcoming scenario is from sentry -> launchpad. The sentry API receives uploads from customer build tooling. Sentry does authentication and request validation and then will spawn a task in launchpad. This will replace the kafka consumer in launchpad.

PR #582, opened 2026-03-31
feat(taskbroker): Fetch Pending Activations in Batches

george-sentry:
## Linear
Completes [STREAM-829](https://linear.app/getsentry/issue/STREAM-829/fetch-tasks-in-batches)

## Description

This PR changes `get_pending_activation` to return multiple of activations (`Vec<InflightActivation>`) instead of a single one (`Option<InflightActivation>`) and to accept a `limit` argument of type `Option<i32>`. In all unmodified areas of the application, that argument is `Some(1)`. In the fetch loop, it is `Some(config.fetch_batch_size)`.

Once a batch of tasks is fetched, they are submitted to the push pool one by one. Some comments...

* We still haven't decided what to do when sending a task fails. Right now, when that happens, the upkeep task will eventually change its status back to pending and increment its processing attempts. Once processing attempts run out, the task will be dropped, which may be a problem. This will be addressed in a future PR
* The new `limit` argument is an optional, but we do not use `None` anywhere in the code, so I'm thinking about making it required and passing in `1` whenever a single activation is required

## Details
- Add a `fetch_batch_size` field to the configuration with a default value of 1
- Change `get_pending_activations` back to `get_pending_activations_from_namespaces` (what it was originally)
- Change `get_pending_activation` to `get_pending_activations`, add a `limit` argument with type `Option<i32>`, and make it return a vector of tasks instead of one
- Change all previously existing calls to `get_pending_activations` to pass `Some(1)` as the limit
- Add three new tests to see what happens when...
  - The limit is `None`
  - The limit is `Some(X)` where X is >= Y where Y is the number of pending tasks
  - The limit is `Some(X)` where X is < Y where Y is the number of pending tasks

### DIFF
--- a/src/grpc/server.rs
+++ b/src/grpc/server.rs
@@ -50,8 +50,8 @@ impl ConsumerService for TaskbrokerServer {
             Ok(activations) if activations.len() > 1 => {
                 error!(
                     count = activations.len(),
-                    application = ?application.as_deref(),
-                    namespace = ?namespace.as_deref(),
+                    application = ?application,
+                    namespace = ?namespace,
                     "get_pending_activations returned more than one row despite limit of 1",
                 );
 
@@ -159,8 +159,8 @@ impl ConsumerService for TaskbrokerServer {
             Ok(activations) if activations.len() > 1 => {
                 error!(
                     count = activations.len(),
-                    application = ?application.as_deref(),
-                    namespace = ?namespace.as_deref(),
+                    application = ?application,
+                    namespace = ?namespace,
                     "get_pending_activations returned more than one row despite limit of 1",
                 );
 


src/store/inflight_activation.rs
```diff
@@ -374,34 +374,33 @@ pub trait InflightActivationStore: Send + Sync {
...
-            return Ok(None);
-        }
-
-        Ok(Some(results[0].clone()))
+        Ok(results)
     }
 
     /// Claim pending activations (moves them to processing), optionally filtered by application and namespaces.
-    async fn get_pending_activations(
+    async fn get_pending_activations_from_namespaces(
```
fpacifici: What was the reason for renaming this ?

PR #586, opened 2026-04-02
feat(taskbroker): Add Claimed Status to Handle Push Failures

george-sentry:
## Linear
Completes [STREAM-860](https://linear.app/getsentry/issue/STREAM-860/add-sending-status-to-handle-push-failures)

## Description

Right now, I rely on processing_deadline to revert processing tasks back to pending if pushing them failed. This isn't good because it eats through processing attempts, resulting in needlessly dropped tasks.

I want to add a `Sending` status that indicates a task is being sent. Now, upkeep increments processing attempts only for tasks that are still in "sending" when their processing deadlines expire. If the status is "processing," that means the task was already sent successfully and its processing attempts can be incremented.

This will help us avoid dropping tasks needlessly when workers are busy.

Note that my original plan was different. You can see it in the commit history. Here is a description of that plan.
> I want to add a sent column to the activation table to track whether a task was successfully sent after being fetched from the table. Now, upkeep increments processing attempts only for tasks that are processing and have sent = true.
> 
> If the status is processing and sent = false, that means pushing failed or timed out (or didn't happen yet), and we can revert back to pending without incrementing processing attempts.

### DIFF
--- a/src/grpc/server.rs
+++ b/src/grpc/server.rs
@@ -137,6 +137,7 @@ impl ConsumerService for TaskbrokerServer {
                 Err(Status::internal("Unable to fetch next task"))
             }
 
+            // If we return an error, the worker will place the result back in its internal queue and send the update again in the future, which is not desired
             Ok(None) => Ok(Response::new(SetTaskStatusResponse { task: None })),
 
             Ok(Some(inflight)) => {
--- a/src/grpc/server.rs
+++ b/src/grpc/server.rs
@@ -11,7 +11,7 @@ use tonic::{Request, Response, Status};
 
 use crate::config::{Config, DeliveryMode};
 use crate::store::inflight_activation::{InflightActivationStatus, InflightActivationStore};
-use tracing::{error, instrument};
+use tracing::{error, instrument, warn};
 
 pub struct TaskbrokerServer {
     pub store: Arc<dyn InflightActivationStore>,
@@ -42,7 +42,7 @@ impl ConsumerService for TaskbrokerServer {
             .await;
 
         match inflight {
-            Ok(None) => Err(Status::not_found("No pending activation")),
+            Ok(None) => Err(Status::not_found("No pending activations")),
 
             Ok(Some(inflight)) => {
                 let now = Utc::now();
@@ -137,8 +137,12 @@ impl ConsumerService for TaskbrokerServer {
                 Err(Status::internal("Unable to fetch next task"))
             }
 
-            // If we return an error, the worker will place the result back in its internal queue and send the update again in the future, which is not desired
-            Ok(None) => Ok(Response::new(SetTaskStatusResponse { task: None })),
+            Ok(None) => {
+                warn!("No pending activations");
+
+                // If we return an error, the worker will place the result back in its internal queue and send the update again in the future, which is not desired
+                Ok(Response::new(SetTaskStatusResponse { task: None }))
+            }
 
             Ok(Some(inflight)) => {
                 if inflight.processing_attempts < 1 {
--- a/src/grpc/server.rs
+++ b/src/grpc/server.rs
@@ -35,31 +35,16 @@ impl ConsumerService for TaskbrokerServer {
 
         let application = &request.get_ref().application;
         let namespace = &request.get_ref().namespace;
-        let namespaces = namespace.as_ref().map(std::slice::from_ref);
 
         let inflight = self
             .store
-            .get_pending_activations(application.as_deref(), namespaces, Some(1), None)
+            .claim_activation_for_pull(application.as_deref(), namespace.as_deref())
             .await;
 
         match inflight {
-            Ok(activations) if activations.is_empty() => {
-                Err(Status::not_found("No pending activation"))
-            }
+            Ok(None) => Err(Status::not_found("No pending activation")),
 
-            Ok(activations) if activations.len() > 1 => {
-                error!(
-                    count = activations.len(),
-                    application = ?application,
-                    namespace = ?namespace,
-                    "get_pending_activations returned more than one row despite limit of 1",
-                );
-
-                Err(Status::internal("Unable to retrieve pending activation"))
-            }
-
-            Ok(activations) => {
-                let inflight = &activations[0];
+            Ok(Some(inflight)) => {
                 let now = Utc::now();
 
                 if inflight.processing_attempts < 1 {
@@ -141,10 +126,10 @@ impl ConsumerService for TaskbrokerServer {
         };
 
         let start_time = Instant::now();
-        let namespaces = namespace.as_ref().map(std::slice::from_ref);
+
         let res = match self
             .store
-            .get_pending_activations(application.as_deref(), namespaces, Some(1), None)
+            .claim_activation_for_pull(application.as_deref(), namespace.as_deref())
             .await
         {
             Err(e) => {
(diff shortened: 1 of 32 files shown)


src/grpc/server.rs
```diff
@@ -141,35 +126,21 @@ impl ConsumerService for TaskbrokerServer {
...
-                    namespace = ?namespace,
-                    "get_pending_activations returned more than one row despite limit of 1",
-                );
-
-                Err(Status::internal("Unable to fetch next task"))
-            }
-
-            Ok(activations) => {
-                let inflight = &activations[0];
+            // If we return an error, the worker will place the result back in its internal queue and send the update again in the future, which is not desired
```
evanh: We should still catch these cases and at least log an error.

src/store/postgres_activation_store.rs
```diff
@@ -513,53 +537,66 @@ impl InflightActivationStore for PostgresActivationStore {
...
+             SET processing_deadline = null,
+                 status = $1
+             WHERE processing_deadline < $2
+                 AND at_most_once = TRUE
+                 AND (status = $3 OR status = $4)",
         )
         .bind(InflightActivationStatus::Failure.to_string())
         .bind(now)
         .bind(InflightActivationStatus::Processing.to_string())
+        .bind(InflightActivationStatus::Sending.to_string())
```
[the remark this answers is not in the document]
george-sentry: This is intentional. It is possible to deliver a task to a worker successfully but observe a send failure due to downstream network issues. For example, the response may not reach the taskbroker before the request times out if the network is congested. If I used the AI's suggestion in this scenario, the task would be mistakenly retried, resulting in more than one execution (a clear violation of the at-most-once policy).

src/grpc/server.rs
george-sentry: Added logging here!

src/store/inflight_activation.rs
```diff
@@ -37,6 +37,7 @@ pub enum InflightActivationStatus {
     /// Unused but necessary to align with sentry-protos
     Unspecified,
     Pending,
+    Sending,
```
markstory: What if we had a `Claimed` status? The lifecycle could be

```mermaid
flowchart

pending -- taken by a push thread and going to send soon --> claimed
claimed -- sent to a worker successfully --> processing
processing --> complete
processing --> retry
processing --> failure
```

You have 'claim' in a bunch of the methods, but no status reflecting that, and with the addition of `Sending` the `Processing` status can mean two different things depending on the broker mode.

src/store/inflight_activation.rs
```diff
@@ -374,15 +400,26 @@ pub trait InflightActivationStore: Send + Sync {
...
+    async fn claim_activations(
+        &self,
+        application: Option<&str>,
+        namespaces: Option<&[String]>,
+        limit: Option<i32>,
+        bucket: Option<BucketRange>,
+        status: InflightActivationStatus,
+    ) -> Result<Vec<InflightActivation>, Error>;
+
+    /// Claims `limit` activations within the `bucket` range. Push mode uses status `Sending` until `mark_activation_sent` moves to `Processing`.
```
markstory: With the methods using `claim` in their name, `Claimed` could be a better status name.

src/store/inflight_activation.rs
```diff
@@ -916,6 +985,31 @@ impl InflightActivationStore for SqliteActivationStore {
         Ok(rows.into_iter().map(|row| row.into()).collect())
     }
 
+    #[instrument(skip_all)]
+    async fn mark_activation_sent(&self, id: &str) -> Result<(), Error> {
```
markstory: Sent isn't a status/state in the state machine, should this be `mark_activation_processing`?

src/store/inflight_activation.rs
```diff
@@ -1060,53 +1154,66 @@ impl InflightActivationStore for SqliteActivationStore {
...
+        .bind(InflightActivationStatus::Sending)
         .execute(&mut *atomic)
-        .await;
+        .await?;
 
-        let mut processing_deadline_modified_rows = 0;
-        if let Ok(query_res) = most_once_result {
-            processing_deadline_modified_rows = query_res.rows_affected();
-        }
+        // Revert activations that weren't delivered back to 'pending' without consuming an attempt
```
markstory: In other tasks systems this is referred to as releasing a claim

src/store/inflight_activation.rs
evanh: This is functionally the same as before, right? Changing `sending` to `claimed`?

george-sentry: It sounds like it is. I don't mind changing `sending` to `claimed` if it's more clear that way.

src/store/inflight_activation.rs
george-sentry: Good point - yes.

src/store/inflight_activation.rs
george-sentry: I think this is a good idea. Thoughts @evanh?

evanh: Yeah I think it makes sense. Cleans up the nomenclature for sure.

src/store/inflight_activation.rs
markstory: Yeah, it is just a naming change.

src/store/postgres_activation_store.rs
```diff
@@ -513,6 +568,50 @@ impl InflightActivationStore for PostgresActivationStore {
...
+             WHERE at_most_once = TRUE
+                 AND status = $2
+                 AND claim_expires_at IS NOT NULL
+                 AND claim_expires_at < $3",
+        )
+        .bind(InflightActivationStatus::Failure.to_string())
+        .bind(InflightActivationStatus::Claimed.to_string())
+        .bind(now)
+        .execute(&mut *atomic)
+        .await?;
```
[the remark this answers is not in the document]
george-sentry: I clearly said in the comment on this block of code that AMO tasks should be moved to failure in this case because a perceived publish failure could be the result of a downstream network fault (response data failing to return back to us). That means if publishing is actually successful but we perceive a failure and revert the task back to pending as a result, we will potentially send the task twice.

But the AI reviewers still have an issue with this. Are they right?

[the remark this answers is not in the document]
markstory: If delivery to a worker fails (due to network issues/timeout) we don't know if the task was actually received by a worker, and moving the task to Failed, could result on claim expiration, could result in the task never being attempted. Given that workers also play a part in making at-most-once idempotent, re-sending the task in this scenario would avoid the task being executed 0 times.

markstory (approved):
Looks good to me, outside of the at-most-once task handling. While I don't think what you have is wrong, it does run the risk of dropping tasks without ever executing them.

src/store/postgres_activation_store.rs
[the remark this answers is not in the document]
george-sentry: So you think it would be wiser to revert AMO tasks back to pending instead of failing them in this case, and run a minor risk of executing them twice?

evanh (approved):

src/grpc/server.rs
```diff
@@ -141,35 +126,25 @@ impl ConsumerService for TaskbrokerServer {
...
-
-            Ok(activations) if activations.len() > 1 => {
-                error!(
-                    count = activations.len(),
-                    application = ?application,
-                    namespace = ?namespace,
-                    "get_pending_activations returned more than one row despite limit of 1",
-                );
+            Ok(None) => {
+                warn!("No pending activations");
```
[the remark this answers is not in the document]
george-sentry: Not a problem.

src/store/inflight_activation.rs
```diff
@@ -340,6 +352,7 @@ impl InflightActivationStoreConfig {
             max_processing_attempts: config.max_processing_attempts,
             vacuum_page_count: config.vacuum_page_count,
             processing_deadline_grace_sec: config.processing_deadline_grace_sec,
+            claim_lease_ms: config.fetch_batch_size.max(1) as u64 * config.push_queue_timeout_ms,
```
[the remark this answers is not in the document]
george-sentry: This is possibly a valid point. It may be a better idea to just make this value configurable, but then we'll need to pick an appropriate value, which is hard. Let's remember this for the future.

src/upkeep.rs
```diff
@@ -299,7 +318,7 @@ pub async fn do_upkeep(
                 .expect("Could not create kafka producer in upkeep"),
         );
         if let Ok(tasks) = store
-            .get_pending_activations_from_namespaces(None, Some(&demoted_namespaces), None, None)
+            .claim_activations(None, Some(&demoted_namespaces), None, None, false)
             .await
         {
             // Produce tasks to Kafka with updated namespace
```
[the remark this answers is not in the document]
george-sentry: The AI reviewers constantly get tripped up by this code. If I mark them as claimed, it will complain about failed publishing resulting in infinite retries. If I mark them as processing, it will complain that we will run out of tries after the processing deadline is exceeded too many times. In other words, no matter which way you go, this chunk of code will be flagged.

PR #587, opened 2026-04-03
Add headers and hooks to taskbroker client

gricha:
Relevant context: https://www.notion.so/sentry/RFC-Unified-ViewerContext-via-ContextVar-32f8b10e4b5d81988625cb5787035e02

I think if we want to implicitly pass VC across the wire, we'll need to do these.

to make this work fully implicitly i add hooks to both pack it into context as header on producer side and unpack it on the consumer side automatically.


george-sentry:
Everything seems okay here, aside from maybe the AI reviewer comments. The only impact I can see is that task activations may be larger, which could result in a small decrease in throughput. I will let someone else review and approve though 🙂

PR #589, opened 2026-04-07
feat(taskbroker): Implement Global Backoff Mechanism

george-sentry:
## Linear
Completes [STREAM-862](https://linear.app/getsentry/issue/STREAM-862/implement-global-backoff-mechanism)

## Description

This PR introduces a dynamic global backoff mechanism to slow down the push rate when workers are busy. Here is the basic idea...
* Every time there is a submit or push failure, we increase the global backoff value up to a configured maximum
* A background task gradually decrements the backoff value until it reaches zero again, or until an event triggers an increase instead - this is the "leaky bucket" part
* Before submitting each activation to the push pool, the fetch thread sleeps for the current backoff wait time, reducing in a smoother increase / decrease in throughput when workers are busy

### DIFF
--- a/src/backoff.rs
+++ b/src/backoff.rs
@@ -1,0 +1,180 @@
+use std::{
+    sync::atomic::{AtomicU32, Ordering},
+    time::Duration,
+};
+
+use crate::config::Config;
+
+/// Shared leaky bucket wait to pace fetching in push mode.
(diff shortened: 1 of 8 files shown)


george-sentry:
Currently not necessary.

PR #592, opened 2026-04-13
ref(eco): Adds report_timeout_errors options to Task definitions, allows ProcessingDeadlineExceeded to be retried

GabeVillalobos:
Updates the python lib code with the following changes:
1. Broadens allowable retry exceptions in `retry.py` to include BaseException types (like `ProcessingDeadlineExceeded`)
2. Updates workerchild to check for retries on the `ProcessingDeadlineExceeded` handling branch.
3. Adds a new `report_timeout_errors` check which allows us to opt out of Sentry reporting on a per-task definition basis.

This will allow developers to optionally specify retries on task timeouts.

Resolves ISWF-2447

### DIFF
--- a/clients/python/src/taskbroker_client/retry.py
+++ b/clients/python/src/taskbroker_client/retry.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from enum import Enum
+from multiprocessing.context import TimeoutError
 
 from sentry_protos.taskbroker.v1.taskbroker_pb2 import (
     ON_ATTEMPTS_EXCEEDED_DEADLETTER,
@@ -68,14 +69,12 @@ class Retry:
         ignore: tuple[type[BaseException], ...] | None = None,
         times_exceeded: LastAction = LastAction.Discard,
         delay: int | None = None,
-        retry_on_timeout: bool = True,
     ):
         self._times = times
         self._allowed_exception_types: tuple[type[BaseException], ...] = on or ()
         self._denied_exception_types: tuple[type[BaseException], ...] = ignore or ()
         self._times_exceeded = times_exceeded
         self._delay = delay
-        self._retry_on_timeout = retry_on_timeout
 
     def max_attempts_reached(self, state: RetryState) -> bool:
         # We subtract one, as attempts starts at 0, but `times`
@@ -95,8 +94,9 @@ class Retry:
         if isinstance(exc, self._denied_exception_types):
             return False
 
-        # Retry for allowed exception types
-        if isinstance(exc, self._allowed_exception_types):
+        # In the retry allow list or processing deadline is exceeded
+        # When processing deadline is exceeded, the subprocess raises a TimeoutError
+        if isinstance(exc, (TimeoutError, self._allowed_exception_types)):
             return True
 
         return False
--- a/clients/python/src/taskbroker_client/retry.py
+++ b/clients/python/src/taskbroker_client/retry.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import logging
 from enum import Enum
-from multiprocessing.context import TimeoutError
 
 from sentry_protos.taskbroker.v1.taskbroker_pb2 import (
     ON_ATTEMPTS_EXCEEDED_DEADLETTER,
@@ -69,19 +68,21 @@ class Retry:
         ignore: tuple[type[BaseException], ...] | None = None,
         times_exceeded: LastAction = LastAction.Discard,
         delay: int | None = None,
+        retry_on_timeout: bool = True,
     ):
         self._times = times
         self._allowed_exception_types: tuple[type[BaseException], ...] = on or ()
         self._denied_exception_types: tuple[type[BaseException], ...] = ignore or ()
         self._times_exceeded = times_exceeded
         self._delay = delay
+        self._retry_on_timeout = retry_on_timeout
 
     def max_attempts_reached(self, state: RetryState) -> bool:
         # We subtract one, as attempts starts at 0, but `times`
         # starts at 1.
         return state.attempts >= (self._times - 1)
 
-    def should_retry(self, state: RetryState, exc: Exception) -> bool:
+    def should_retry(self, state: RetryState, exc: BaseException) -> bool:
         # If there are no retries remaining we should not retry
         if self.max_attempts_reached(state):
             return False
@@ -94,9 +95,13 @@ class Retry:
         if isinstance(exc, self._denied_exception_types):
             return False
 
-        # In the retry allow list or processing deadline is exceeded
-        # When processing deadline is exceeded, the subprocess raises a TimeoutError
-        if isinstance(exc, (TimeoutError, self._allowed_exception_types)):
+        # Retry for allowed exception types
+        if isinstance(exc, self._allowed_exception_types):
+            return True
+
+        from taskbroker_client.worker.workerchild import ProcessingDeadlineExceeded
+
+        if self._retry_on_timeout and isinstance(exc, ProcessingDeadlineExceeded):
             return True
 
         return False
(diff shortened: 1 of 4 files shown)


clients/python/src/taskbroker_client/retry.py
```diff
@@ -94,9 +95,9 @@ def should_retry(self, state: RetryState, exc: Exception) -> bool:
         if isinstance(exc, self._denied_exception_types):
             return False
 
-        # In the retry allow list or processing deadline is exceeded
-        # When processing deadline is exceeded, the subprocess raises a TimeoutError
-        if isinstance(exc, (TimeoutError, self._allowed_exception_types)):
```
GabeVillalobos, line 99: This is incorrect, I need to re-add allowed exc types.

clients/python/src/taskbroker_client/retry.py
```diff
@@ -69,19 +68,21 @@ def __init__(
         ignore: tuple[type[BaseException], ...] | None = None,
         times_exceeded: LastAction = LastAction.Discard,
         delay: int | None = None,
+        retry_on_timeout: bool = True,
```
kcons: given that we _can_ specify `ProcessingDeadlineException` in `on` already and with your change it can work, `retry_on_timeout` seems not entirely orthogonal. here.

I'm interested in why we use `type[BaseException]` rather than `type[Exception]`, what with `should_retry` (at least prior to this) only allowing `Exception`. I can see an argument for it, but not the motivating case.

If we try to make `retry_on_timeout` stand alone, we potentially don't need to involve `should_retry` at all or do that localized import, we just check the bool. This seems oddly consistent with how task timeouts aren't the same thing as exceptions that escape; it's framework functionality that may or may not be an exception.

GabeVillalobos: > given that we can specify ProcessingDeadlineException in on already and with your change it can work, retry_on_timeout seems not entirely orthogonal. here.

I agree, I think we _could_ fold this in, but it seemed distinct enough at first with the special handling this case currently has. Perhaps there's enough here to roll the exception handling in `worker_child` together with the generic exception handling to simplify all of this though.

> I'm interested in why we use type[BaseException] rather than type[Exception], what with should_retry (at least prior to this) only allowing Exception. I can see an argument for it, but not the motivating case.

I am as well.

I'll let the taskbroker team weigh in on this since they're more likely to have a strong opinion on this.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -245,7 +246,11 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
                         "taskname": inflight.activation.taskname,
                     },
                 )
-                next_state = TASK_ACTIVATION_STATUS_FAILURE
+                retry = task_func.retry
+                if retry and retry.should_retry(inflight.activation.retry_state, err):
```
markstory, line 250: Would you have tasks configured to do retries on `ProcessingDeadlineExceeded`?

clients/python/src/taskbroker_client/retry.py
markstory: > I'm interested in why we use type[BaseException] rather than type[Exception], what with should_retry (at least prior to this) only allowing Exception. I can see an argument for it, but not the motivating case.

I don't have a strong reason to keep `BaseException`. We could easily go to `Exception` and be compatible with application usage.

> given that we can specify ProcessingDeadlineException in on already and with your change it can work, 

Including `ProcessingDeadlineException` in an `on` clause works for me. It exposes less API surface area, and makes it more explicit on what kinds of timeouts are captured as the `retry_on_timeout` parameter doesn't explain which kinds of timeouts much.

clients/python/src/taskbroker_client/retry.py
```diff
@@ -94,9 +95,13 @@ def should_retry(self, state: RetryState, exc: Exception) -> bool:
...
             return False
 
-        # In the retry allow list or processing deadline is exceeded
-        # When processing deadline is exceeded, the subprocess raises a TimeoutError
-        if isinstance(exc, (TimeoutError, self._allowed_exception_types)):
+        # Retry for allowed exception types
+        if isinstance(exc, self._allowed_exception_types):
+            return True
+
+        from taskbroker_client.worker.workerchild import ProcessingDeadlineExceeded
```
markstory: We could move `ProcessingDeadlineExceeded` to `taskbroker_client.types` to break the import cycle.

clients/python/src/taskbroker_client/retry.py
```diff
@@ -94,9 +95,13 @@ def should_retry(self, state: RetryState, exc: Exception) -> bool:
         if isinstance(exc, self._denied_exception_types):
             return False
 
-        # In the retry allow list or processing deadline is exceeded
-        # When processing deadline is exceeded, the subprocess raises a TimeoutError
-        if isinstance(exc, (TimeoutError, self._allowed_exception_types)):
```
markstory, line 99: I don't think this is a safe change. If tasks did have `TimeoutError` being raised it would previously automatically retry and now the task will not retry.

GabeVillalobos: Where is TimeoutError being raised in the code? I was a bit confused on where this is actually going to be an issue since I only saw the `ProcessingDeadlineExceeded` exception being raised here: https://github.com/getsentry/taskbroker/blob/b7f98fc4b782d4ed92a72d5fe44290cf899bed09/clients/python/src/taskbroker_client/worker/workerchild.py#L228

clients/python/src/taskbroker_client/worker/workerchild.py
GabeVillalobos: The use case for us would be: we try a webhook dispatch, it fails with a timeout, so we retry x number of times total before bailing on the webhook attempt entirely. If there's another way to achieve this without hooking into processingDeadlineExceeded except block here, I'm game for that.

clients/python/src/taskbroker_client/retry.py
GabeVillalobos: I ended up just broadening the allowable exception types to include `BaseException` in `should_retry`. To me, it makes sense for `ProcessingDeadline` to continue to be categorized as a `BaseException` type, since it's not strictly an error, but a system level timer instead.

clients/python/src/taskbroker_client/retry.py
GabeVillalobos: Restored this in case it breaks anything, but I do think there's a case to deprecate this behavior, I just don't want to regress any behavior in the meantime.

clients/python/src/taskbroker_client/worker/workerchild.py
markstory: You could have the task processing deadline be longer than the cumulative timeouts of your HTTP request  (and retries). That would let you do some retries within the task, handle those failures, and request a full task retry without having to catch processing deadlines.

clients/python/src/taskbroker_client/retry.py
```diff
@@ -81,7 +81,7 @@ def max_attempts_reached(self, state: RetryState) -> bool:
         # starts at 1.
         return state.attempts >= (self._times - 1)
 
-    def should_retry(self, state: RetryState, exc: Exception) -> bool:
+    def should_retry(self, state: RetryState, exc: BaseException) -> bool:
         # If there are no retries remaining we should not retry
         if self.max_attempts_reached(state):
             return False
```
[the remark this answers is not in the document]
markstory: I think that's what we want now.

markstory (approved):
Looks good to me.  Once this is merged and used in sentry, it would be good to remove the deadline exception workarounds we have in sentry.

clients/python/src/taskbroker_client/worker/workerchild.py
GabeVillalobos: Discussed off issue but the issue we're hitting is with the DNS resolution in certain cases (ngrok is typically a culprit here), which we can't meaningfully apply a timeout to. This was a way for us to allow retries without reimplementing the same logic we already have here.

PR #593, opened 2026-04-14
feat: Separate push Taskworker from pull Taskworker

evanh:
Moving the push taskbroker to a k8s Deployment means that the worker logic can simplify. Particularly the push worker doesn't need a list of hosts, just a single service host. Instead of having one Taskworker that supports both modes, create separate classes and clients for each.

This meant moving the multiprocess part of the TaskWorker to a separate class, that manages the queues and the various children processes and threads. This class is shared by both the Push worker and the original pull worker.
The push-based taskworker doesn't need all the features of the pull based taskbroker. Create a new simpler client that has a single RPC host to connect to, that the AlloyDB workers can use.

Note that this should be a non-breaking change: the is being moved around, not changed.


clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -486,3 +486,105 @@ def update_task(
...
+        except grpc.RpcError as err:
+            self._metrics.incr(
+                "taskworker.client.rpc_error",
+                tags={"method": "SetTaskStatus", "status": err.code().name},
+            )
+            if err.code() == grpc.StatusCode.NOT_FOUND:
+                # The task was not found, so we can't update it.
+                return False
+            if err.code() == grpc.StatusCode.UNAVAILABLE:
+                # The brokers are not responding, so we can't update the task.
```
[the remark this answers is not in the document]
george-sentry: Seems like a real problem. Since this method is called inside of a `try ... except` and its return value is unused, maybe it makes sense to remove the return value altogether and mention in a comment that the function raises an exception when updating fails and in that case the update should be retried?

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -490,33 +580,76 @@ def spawn_children_thread() -> None:
...
+        logger.info("taskworker.worker.shutdown.result")
+        if self._result_thread:
+            # Use a timeout as sometimes this thread can deadlock on the Event.
+            self._result_thread.join(timeout=5)
+
+        # Drain any remaining results synchronously
+        while True:
+            try:
+                result = self._processed_tasks.get_nowait()
+                self._send_result(result, True)
```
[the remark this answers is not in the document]
george-sentry: This is probably okay since the worker is shutting down, but I guess it wouldn't hurt to address with something like this.

```py
# Drain any remaining results synchronously
while True:
    try:
        result = self._processed_tasks.get_nowait()
        self._send_result(result, True)
    except queue.Empty:
        break
    except RequeueException:
        break
```

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -131,44 +137,71 @@ def __init__(
...
+                "processing_pool": self._processing_pool_name,
+            },
+        )
+        # Use the shutdown_event as a sleep mechanism
+        self._grpc_sync_event.wait(self._setstatus_backoff_seconds)
+
+        try:
+            self.client.update_task(result)
+            self._setstatus_backoff_seconds = 0
+            return None
```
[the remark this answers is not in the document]
george-sentry: Sentry bot caught this one too.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -426,21 +395,20 @@ def _send_update_task(
             },
         )
         # Use the shutdown_event as a sleep mechanism
-        self._shutdown_event.wait(self._setstatus_backoff_seconds)
+        self._grpc_sync_event.wait(self._setstatus_backoff_seconds)
```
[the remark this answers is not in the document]
george-sentry: Could be an issue, not sure. What does this change do?

clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -486,3 +486,105 @@ def update_task(
...
+        self._application = application
+        self._service = service
+        self._rpc_secret = rpc_secret
+        self._metrics = metrics
+
+        self._grpc_options: list[tuple[str, Any]] = [
+            ("grpc.max_receive_message_length", MAX_ACTIVATION_SIZE)
+        ]
+        if grpc_config:
+            self._grpc_options.append(("grpc.service_config", grpc_config))
```
markstory, line 517: I don't know if we've ever used this in the existing worker either.

clients/python/src/taskbroker_client/worker/client.py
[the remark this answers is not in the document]
markstory: I agree, a failed update needs to be retried until it succeeds or the worker is killed.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -178,238 +211,174 @@ def signal_handler(*args: Any) -> None:
...
-            child.join()
+class TaskWorker:
+    """
+    A TaskWorker fetches tasks from a taskworker RPC host and handles executing task activations.
 
-        logger.info("taskworker.worker.shutdown.result")
-        if self._result_thread:
-            # Use a timeout as sometimes this thread can deadlock on the Event.
-            self._result_thread.join(timeout=5)
+    Tasks are executed in a forked process so that processing timeouts can be enforced.
```
markstory:
```suggestion
    Tasks are executed in a forked/spawned process so that processing timeouts can be enforced.
```

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -178,238 +211,174 @@ def signal_handler(*args: Any) -> None:
...
+    will be fetched.
 
-        # Drain any remaining results synchronously
-        while True:
-            try:
-                result = self._processed_tasks.get_nowait()
-                self._send_result(result, fetch=False)
-            except queue.Empty:
-                break
+    Taskworkers can be run with `sentry run taskworker`
```
markstory: I guess we don't need this here anymore.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -178,238 +211,174 @@ def signal_handler(*args: Any) -> None:
...
-            finally:
-                if server is not None:
-                    server.stop(grace=5)
-
-                self.shutdown()
-        else:
-            try:
-                while True:
-                    self.run_once()
+                server.wait_for_termination()
```
markstory, line 271: Will this handle both `SIGINT` and `SIGTERM`?

clients/python/src/taskbroker_client/worker/worker.py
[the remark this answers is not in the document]
markstory: Ideally we would retry sending a couple of times to reduce the number of tasks that are executed multiple times.

clients/python/src/taskbroker_client/worker/client.py
evanh: I'll leave it for continuity.

clients/python/src/taskbroker_client/worker/worker.py
evanh: Update to handle that. Good comment.

clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -486,3 +486,114 @@ def update_task(
                 receive_timestamp=time.monotonic(),
             )
         return None
+
+
+class PushTaskbrokerClient:
+    """
+    Taskworker RPC client wrapper
+
+    Push brokers are a deployment so don't need to be connected to individually. There is one service provided
```
markstory:
```suggestion
    Push brokers are a deployment so they don't need to be connected to individually. There is one service provided
```

clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -486,3 +486,114 @@ def update_task(
...
+        return ConsumerServiceStub(channel)
+
+    def update_task(
+        self,
+        processing_result: ProcessingResult,
+    ) -> bool:
+        """
+        Update the status for a given task activation.
+
+        The return value is the next task that should be executed.
```
markstory:
```suggestion
        Update the status for a given task activation.
```

markstory (approved):
Looks good to me.

PR #594, opened 2026-04-14
ref(store): Restructure Store Module

george-sentry:
## Linear
Completes [STREAM-875](https://linear.app/getsentry/issue/STREAM-875/restructure-store-module)

## Description
The store module is not organized very well right now. It's basically centered around one large file, `inflight_activation.rs`, which mixes together types representing activations, shared helper types, the main store trait, and the SQLite adapter implementation. Meanwhile, the Postgres implementation lives separately in `posgres_activation_store.rs`.

We need a new layout that makes responsibilities explicit. There should be separate files for core activation types, shared supporting types, traits, adapters, and tests.

## Changes
* Move store trait into `store/traits.rs` (plural because I plan to break _this_ apart into smaller traits in the future)
* Move activation and activation status types into `store/activation.rs`
* Move shared supporting types like `BucketRange` and `DepthCounts` into `store/types.rs`
* Move Postgres and SQLite implementations into `store/adapters` module
* Keep tests in `store/tests.rs`
* Update all imports across the codebase to reference these changes

Note that **there should be no changes to behavior.** This is mostly just moving code around.

### DIFF
--- a/benches/store_bench.rs
+++ b/benches/store_bench.rs
@@ -4,10 +4,9 @@ use chrono::Utc;
 use criterion::{Criterion, criterion_group, criterion_main};
 use rand::Rng;
 use taskbroker::{
-    store::inflight_activation::{
-        InflightActivationStatus, InflightActivationStore, InflightActivationStoreConfig,
-        SqliteActivationStore,
-    },
+    store::activation::InflightActivationStatus,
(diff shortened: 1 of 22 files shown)


markstory (approved):
Looking good

PR #600, opened 2026-04-16
ref(taskbroker): Flatten Imports

george-sentry:
## Linear
Completes [STREAM-878](https://linear.app/getsentry/issue/STREAM-878/flatten-imports)
Completes [STREAM-885](https://linear.app/getsentry/issue/STREAM-885/add-agentsmd-file)

## Description
Unfortunately, `rustfmt` combined with autocompleted imports tends to results in massive unreadable `use` statements with multiple layers of nesting, such as this piece of work at the top of main.rs.

```rs
use taskbroker::kafka::{
    admin::create_missing_topics,
    consumer::start_consumer,
    deserialize_activation::{self, DeserializeActivationConfig},
    inflight_activation_writer::{ActivationWriterConfig, InflightActivationWriter},
    os_stream_writer::{OsStream, OsStreamWriter},
};
```

Not very readable. Let's "flatten" these imports to just a single level of nesting.

```rs
use taskbroker::kafka::admin::create_missing_topics;
use taskbroker::kafka::consumer::start_consumer;
use taskbroker::kafka::deserialize_activation::{self, DeserializeActivationConfig};
use taskbroker::kafka::inflight_activation_writer::{ActivationWriterConfig, InflightActivationWriter};
use taskbroker::kafka::os_stream_writer::{OsStream, OsStreamWriter};
```

It's easier to read when every use statement is exactly one line, I think. Many popular Rust projects use this approach, such as `serde`. Here is an example from that crate.

```rs
use crate::internals::ast::{Container, Data, Field, Style};
use crate::internals::attr::{Default, Identifier, TagType};
use crate::internals::{ungroup, Ctxt, Derive};
...
```

It will be better on everyone's eyes to make this change, and it will also bring our codebase formatting closer to the formatting in real and widely used Rust projects.

**There should be no changes to behavior.**

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,65 @@
+# Taskbroker Development Guide for Agents
+
+## Overview
+Taskbroker is Sentry's distributed asynchronous task execution system. It consists of taskworkers (an application that executes tasks) and taskbrokers (an application that distributes tasks across workers).
+
+## Formatting
+Run `make format` before you commit. There are some other conventions you should follow that aren't enforced by that command, described below.
+
+### Rust Imports
+
+#### Flatten Imports
+Do not nest braced groups inside other braced groups. Use only one level of nesting per `use` statement. Here is an example.
+
+```rs
+use a::{ b::{c, d}, e::{f, g} }
+```
+
+Breaking apart this single statement into two improves readability.
+
+```rs
+use a::b::{c, d};
+use a::e::{f, g};
+```
+
+#### Group Imports
+
+Grouping is optional if there are **fewer than four** `use` lines. If there are **four or more** use lines, separate groups with a blank line in this order.
+
+  1. Standard library (`std::`)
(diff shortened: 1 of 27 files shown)


markstory:
Looks good. Is there a way to convince `rustfmt` / LSP autoimport to play nicely with these import groupings?

markstory (approved):

george-sentry:
> Looks good. Is there a way to convince `rustfmt` / LSP autoimport to play nicely with these import groupings?

Unfortunately no. There is an unstable `rustfmt` feature to control import granularity, but I don't think it will be stable any time soon (see [this](https://github.com/rust-lang/rustfmt/issues/4991) issue). As for the grouping / spacing, that doesn't really matter so it's okay if it's not followed.

If the import grouping is too annoying for people / AI we can drop it, just thought it would be fun to try.

PR #604, opened 2026-04-20
feat(taskbroker): Add Application to Worker Pool Mapping

george-sentry:
## Linear
Completes [STREAM-886](https://linear.app/getsentry/issue/STREAM-886/add-application-to-worker-pool-mapping)

## Description

This PR adds a mapping from applications to worker endpoints to the taskbroker configuration. In pull mode, workers from different "applications" (also called "external" workers) simply pull the correct tasks from the broker by specifying the application they need. In push mode, we need to specify the correct worker endpoint for every possible application instead.

Specific changes...
* Removes `worker_endpoint`, `application`, and `namespace` configuration fields
* Adds `worker_map` configuration field with default `getsentry ↦ http://127.0.0.1:50052`
* Modifies configuration unit tests to reflect additions and removals
* Modifies `PushPool::start` to give each push thread a map from applications to worker connections rather than a single worker connection
* Modifies push threads to log an error if a task's application field does not exist in the mapping
* Removes application and namespace arguments from `claim_activations_for_push` as we now want to fetch ALL applications and ALL namespaces, and route based on application


src/push/mod.rs
```diff
@@ -165,8 +173,18 @@ impl PushPool {
...
 
+                                let Some(worker) = workers.get_mut(&activation.application) else {
+                                    error!(
+                                        task_id = %id,
+                                        application = activation.application,
+                                        "Task application has no worker pool mapping"
+                                    );
+
+                                    continue
+                                };
```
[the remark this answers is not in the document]
george-sentry: Sounds legit. What do we want to do in this case? This can only happen due to misconfiguration, and in that case I think it's OK to loop between `pending` and `claimed` until the misconfiguration is addressed.

Here's how this could happen IRL, and the steps we would take to address it.
1. Someone forgets to add an application → worker pool mapping to the taskbroker config _or_ someone routes tasks from a new application into an existing topic
2. Taskbroker claims task (moves from `pending` to `claimed`) but cannot push because there is no mapping for its application
3. We get an alert because we increment a counter every time this happens
4. We add the missing mapping to the taskbroker configuration
5. Task eventually moves back to `pending` and is eventually claimed again
6. This time, there is a mapping for the application and the push is successful
7. If misconfiguration was due to missing mapping, we are done. If it's because task was routed to wrong topic, we fix the routing and wait until no more tasks are being sent to this processing pool
8. Once no more tasks are being sent to this processing pool, we can safely remove the mapping for the application from the broker and all is back to normal

That's if we want to minimize dropping tasks. If we are OK with dropping tasks, we can just use the task killswitch feature. Long story short, I think it's OK to move between `pending` and `claimed` here as long as we can detect that it's happening and we know what to do when it does happen.

Thoughts?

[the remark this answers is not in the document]
markstory: In this scenario, those tasks would also have increasing latency which would eventually alert us, and we could then triage and resolve the configuration gap.

src/push/mod.rs
```diff
@@ -24,7 +25,7 @@ type HmacSha256 = Hmac<Sha256>;
 /// gRPC path for `WorkerService::PushTask` — keep in sync with `sentry_protos` generated client.
 const WORKER_PUSH_TASK_PATH: &str = "/sentry_protos.taskbroker.v1.WorkerService/PushTask";
 
-/// HMAC-SHA256(secret, grpc_path + ":" + message), hex-encoded. Matches Python `RequestSignatureInterceptor` and broker [`crate::grpc::auth_middleware`].
+/// HMAC-SHA256(secret, grpc_path + ":" + message), hex-encoded. Matches Python `RequestSignatureInterceptor` and broker `crate::grpc::auth_middleware`.
```
markstory: Don't the `[]` make links?

src/push/mod.rs
```diff
@@ -130,21 +131,28 @@ impl PushPool {
...
 
-                        Err(e) => {
-                            metrics::counter!("push.worker.connect", "result" => "error")
-                                .increment(1);
-                            error!("Failed to connect to worker - {:?}", e);
+                    for (application, endpoint) in worker_map.clone() {
+                        let worker = match WorkerServiceClient::connect(endpoint).await {
+                            Ok(w) => {
+                                metrics::counter!("push.worker.connect", "result" => "ok")
+                                    .increment(1);
```
markstory, line 140: Could we have a tag for which application worker pool we connected to?

src/push/mod.rs
```diff
@@ -130,21 +131,28 @@ impl PushPool {
...
+                                w
+                            }
 
-                            // When we fail to connect, the taskbroker will shut down, but this may change in the future
-                            return Err(e.into());
-                        }
-                    };
+                            Err(e) => {
+                                metrics::counter!("push.worker.connect", "result" => "error")
```
markstory: Same here for a tag on which application is sad.

src/push/mod.rs
```diff
@@ -165,8 +173,20 @@ impl PushPool {
                                 let id = activation.id.clone();
                                 let callback_url = callback_url.clone();
 
+                                let Some(worker) = workers.get_mut(&activation.application) else {
+                                    metrics::counter!("push.missing_worker_mapping").increment(1);
```
markstory: If we had an application tag here, we would be able to triage broken config faster.

src/config.rs
```diff
@@ -384,11 +378,9 @@ impl Default for Config {
             push_queue_size: 1,
             push_queue_timeout_ms: 5000,
             push_timeout_ms: 30000,
-            worker_endpoint: "http://127.0.0.1:50052".into(),
             callback_addr: "0.0.0.0".into(),
             callback_port: 50051,
-            application: None,
-            namespaces: None,
+            worker_map: [("getsentry".into(), "http://127.0.0.1:50052".into())].into(),
```
markstory:
```suggestion
            worker_map: [("sentry".into(), "http://127.0.0.1:50052".into())].into(),
```

Sentry uses `sentry`.

src/config.rs
```diff
@@ -522,6 +514,10 @@ mod tests {
         assert_eq!(config.max_pending_count, 2048);
         assert_eq!(config.max_processing_count, 2048);
         assert_eq!(config.vacuum_page_count, None);
+        assert_eq!(
+            config.worker_map.get("getsentry").map(String::as_str),
```
markstory:
```suggestion
            config.worker_map.get("sentry").map(String::as_str),
```

src/config.rs
```diff
@@ -549,6 +545,9 @@ mod tests {
                 max_processing_attempts: 5
                 vacuum_page_count: 1000
                 full_vacuum_on_start: true
+                worker_map:
+                    getsentry: http://worker-getsentry:50052
```
markstory:
```suggestion
                    sentry: http://worker-getsentry:50052
```

src/config.rs
```diff
@@ -583,6 +582,19 @@ mod tests {
             assert_eq!(config.vacuum_page_count, Some(1000));
             assert_eq!(config.db_max_size, Some(3_000_000_000));
             assert!(config.full_vacuum_on_start);
+            assert_eq!(
+                config.worker_map,
+                BTreeMap::from([
+                    (
+                        "getsentry".to_owned(),
```
markstory:
```suggestion
                        "sentry".to_owned(),
```

src/config.rs
```diff
@@ -627,6 +639,35 @@ mod tests {
                 config.default_metrics_tags,
                 BTreeMap::from([("key".to_owned(), "value".to_owned())])
             );
+            assert_eq!(
+                config.worker_map.get("getsentry").map(String::as_str),
```
markstory:
```suggestion
                config.worker_map.get("sentry").map(String::as_str),
```

src/config.rs
```diff
@@ -627,6 +639,35 @@ mod tests {
...
+    }
+
+    /// `worker_map` uses the same env map encoding as `default_metrics_tags` (brace `key=value` pairs).
+    #[test]
+    fn test_worker_map_from_env() {
+        Jail::expect_with(|jail| {
+            jail.set_env("TASKBROKER_LOG_FILTER", "error");
+            jail.set_env(
+                "TASKBROKER_WORKER_MAP",
+                "{getsentry=http://127.0.0.1:60052,launchpad=http://127.0.0.1:60053}",
```
markstory:
```suggestion
                "{sentry=http://127.0.0.1:60052,launchpad=http://127.0.0.1:60053}",
```

src/config.rs
```diff
@@ -627,6 +639,35 @@ mod tests {
...
+            );
+
+            let args = Args { config: None };
+            let config = Config::from_args(&args).unwrap();
+            assert_eq!(
+                config.worker_map,
+                BTreeMap::from([
+                    ("getsentry".to_owned(), "http://127.0.0.1:60052".to_owned(),),
```
markstory:
```suggestion
                    ("sentry".to_owned(), "http://127.0.0.1:60052".to_owned(),),
```

src/push/mod.rs
```diff
@@ -209,8 +229,20 @@ impl PushPool {
                         let id = activation.id.clone();
                         let callback_url = callback_url.clone();
 
+                        let Some(worker) = workers.get_mut(&activation.application) else {
+                            metrics::counter!("push.missing_worker_mapping").increment(1);
```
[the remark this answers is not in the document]
markstory: Its ok, we'll get an error in sentry.

markstory (approved):
Looks good to me.

src/push/mod.rs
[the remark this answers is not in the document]
evanh: I think we should have a default option. That way we can have configuration like:

`{"launchpad": "launchpad.com", "default": "regularpool.com"}`

That way when new applications are created they don't need to be explicitly added to the config.

src/push/mod.rs
george-sentry: Yeah, added it back.

src/push/mod.rs
george-sentry: Done!

src/push/mod.rs
george-sentry: Done!

src/config.rs
george-sentry: Fixed all of these.

PR #605, opened 2026-04-22
feat: Add volatile schedule storage

markstory:
Seer has been running without durable storage for scheduled tasks, and it also does not have a redis instance provisioned. With seer's scheduled tasks all using `crontab` expressions, adding a durable storage is very low value, as durable storage is only required for `timedelta` schedules.

Having a volatile storage backend will make it simpler and cheaper to shift seer to taskbroker.

### DIFF
--- a/clients/python/src/taskbroker_client/scheduler/storage.py
+++ b/clients/python/src/taskbroker_client/scheduler/storage.py
@@ -17,34 +17,34 @@ class RunStorageProtocol(Protocol):
     in the future, or adapt taskworkers for other applications should we need to.
     """
 
-    def set(self, key: str, next_runtime: datetime) -> bool: ...
+    def set(self, key: str, next_runtime: datetime) -> bool:
+        """
+        Record a spawn time for a task.
+        The next_runtime parameter indicates when the record should expire,
+        and a task can be spawned again.
 
-    """
-    Record a spawn time for a task.
-    The next_runtime parameter indicates when the record should expire,
-    and a task can be spawned again.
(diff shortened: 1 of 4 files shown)


clients/python/tests/scheduler/test_storage.py
```diff
@@ -0,0 +1,56 @@
+from datetime import UTC, datetime
+
+from taskbroker_client.scheduler.storage import VolatileRunStorage
+
+from ..conftest import freeze_time
+
+
+def test_volatile_run_storage_set_returns_true() -> None:
+    storage = VolatileRunStorage()
```
fpacifici, line 9: Is the redis version covered by the previous RunStorage tests ?

fpacifici (approved):

clients/python/tests/scheduler/test_storage.py
markstory: Yes, all the existing scheduler tests use Redis storage currently.

clients/python/src/taskbroker_client/scheduler/storage.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def read(self, key: str) -> datetime | None:
+        """
+        Retrieve the last run time of a task
+        Returns None if last run time has expired or is unknown.
+        """
+        value = self._data.get(key, None)
+        if value is None:
+            return None
+        return value[0]
```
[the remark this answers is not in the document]
markstory: That's ok. The if the time is in the past, the schedules will still advance.

PR #608, opened 2026-04-27
feat(iswf): Adds silenced_exceptions parameter to tasks, exposes this and report_timeout_errors in task registration

GabeVillalobos:
Exposes the previously added `report_timeout_errors` via task registration method.

Adds a new `silenced_exceptions` parameter, which _should_ put us at feature parity with the remainder of fields in `@retry`:

    | Parameter          | Retry | Report | Raise | Description |
    |--------------------|-------|--------|-------|-------------|
    | on                 | Yes   | Yes    | No    | Exceptions that will trigger a retry & report to Sentry. |
    | on_silent          | Yes   | No     | No    | Exceptions that will trigger a retry but not be captured to Sentry. |
    | exclude            | No    | No     | Yes   | Exceptions that will not trigger a retry and will be raised. |
    | ignore             | No    | No     | No    | Exceptions that will be ignored and not trigger a retry & not report to Sentry. |
    | ignore_and_capture | No    | Yes    | No    | Exceptions that will not trigger a retry and will be captured to Sentry. |

To emulate each of these previous parameters:
**on**
```python
namespace.register(
    ...,
    retry=Retry(times=3, on=(MyCoolException,)),
)
```

**on_silent**
```python
namespace.register(
    ...,
    retry=Retry(times=3, on=(MyCoolException,)),
    silenced_exceptions(MyCoolException,)
)
```

**exclude**
```python
namespace.register(
    ...,
    retry=Retry(times=3, on=(...), ignore=(IgnoreMeException,)),
)
```

**ignore**
```python
namespace.register(
    ...,
    retry=Retry(times=3, ignore=(IgnoreMeException,)),
    silenced_execptions(IgnoreMeException,)
)
```

**ignore_and_capture**
.... It's actually not entirely clear to me how this is supposed to behave compared to `exclude`

### DIFF
--- a/clients/python/src/taskbroker_client/registry.py
+++ b/clients/python/src/taskbroker_client/registry.py
@@ -89,7 +89,7 @@ class TaskNamespace:
         wait_for_delivery: bool = False,
         compression_type: CompressionType = CompressionType.PLAINTEXT,
         report_timeout_errors: bool = True,
-        exceptions_to_silence: tuple[type[BaseException], ...] | None = None,
+        expected_exceptions: tuple[type[BaseException], ...] | None = None,
     ) -> Callable[[Callable[P, R]], Task[P, R]]:
         """
         Register a task.
@@ -117,6 +117,10 @@ class TaskNamespace:
             before returning.
         compression_type: CompressionType
             The compression type to use to compress the task parameters.
+        report_timeout_errors: bool
+            Enable reporting of ProcessingDeadlineExceededError to Sentry.
+        expected_exceptions: tuple[type[BaseException], ...] | None
+            A tuple of exception types that will not be reported by Sentry.
         """
 
         def wrapped(func: Callable[P, R]) -> Task[P, R]:
@@ -136,7 +140,7 @@ class TaskNamespace:
                 wait_for_delivery=wait_for_delivery,
                 compression_type=compression_type,
                 report_timeout_errors=report_timeout_errors,
-                exceptions_to_silence=exceptions_to_silence,
+                expected_exceptions=expected_exceptions,
             )
             # TODO(taskworker) tasks should be registered into the registry
             # so that we can ensure task names are globally unique
@@ -229,7 +233,7 @@ class ExternalNamespace(TaskNamespace):
         wait_for_delivery: bool = False,
         compression_type: CompressionType = CompressionType.PLAINTEXT,
         report_timeout_errors: bool = True,
-        exceptions_to_silence: tuple[type[BaseException], ...] | None = None,
+        expected_exceptions: tuple[type[BaseException], ...] | None = None,
     ) -> Callable[[Callable[P, R]], ExternalTask[P, R]]:
         """
         Register an external task stub.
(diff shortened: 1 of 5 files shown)


clients/python/src/taskbroker_client/registry.py
```diff
@@ -88,6 +88,8 @@ def register(
         at_most_once: bool = False,
         wait_for_delivery: bool = False,
         compression_type: CompressionType = CompressionType.PLAINTEXT,
+        report_timeout_errors: bool = True,
+        exceptions_to_silence: tuple[type[BaseException], ...] | None = None,
```
GabeVillalobos: This name is a little verbose, but the intention here is just to silence sentry errors for some list of exceptions. It should give us the blanket coverage to phase out the remaining options in our retry decorators.

markstory:
> ignore_and_capture

Can this state just cease to exist? There is a single usage of it in sentry.

clients/python/src/taskbroker_client/registry.py
markstory: What about `expected_errors`?  These are errors we expect to happen and don't need observability into.

markstory: Could these new parameters be added to the docstring as well?

GabeVillalobos: Good call, just renamed this and added docstring comments.

GabeVillalobos:
> Can this state just cease to exist? There is a single usage of it in sentry.

@markstory Yeah we can. I think this change still make sense, excluding that use case, since we won't always want to be notified if we retry a retriable exception.

evanh (commented):
Could you add some tests for this behaviour?

GabeVillalobos:
@evanh Added some testing. I think this should cover the major use cases we care about, but can add more if needed.

markstory (approved):

fpacifici:
> | ignore             | No    | No     | No    | Exceptions that will be ignored and not trigger a retry & not report to Sentry. |

What's a real world scenario for letting a task raise an exception, do nothing about that and consider it successful ? 
Shouldn't the task not raise an exception in this case ?

clients/python/src/taskbroker_client/registry.py
```diff
@@ -88,6 +88,8 @@ def register(
         at_most_once: bool = False,
         wait_for_delivery: bool = False,
         compression_type: CompressionType = CompressionType.PLAINTEXT,
+        report_timeout_errors: bool = True,
+        expected_exceptions: tuple[type[BaseException], ...] | None = None,
```
fpacifici: Isn't this naming a bit of a contradiction? "expected exception"
What about `silenced_exceptions`

GabeVillalobos: The original name for this was `exceptions_to_silence` 😅 I do think this is a bit more appropriate. `expected` is a bit vague in what it's implying. I'll update the name again.

GabeVillalobos:
> What's a real world scenario for letting a task raise an exception, do nothing about that and consider it successful ?
Shouldn't the task not raise an exception in this case ?

@fpacifici I largely agree. This seems to be a task definition shorthand to avoiding writing a top-level try/catch, though this PR doesn't introduce this behavior. If we want to deprecate this in favor of slightly more verbose task handlers, I'm in favor of doing this in a separate PR.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -266,6 +266,7 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
             except Exception as err:
                 retry = task_func.retry
                 captured_error = False
+                should_capture_error = not isinstance(err, task_func.silenced_exceptions)
                 if retry:
                     if retry.should_retry(inflight.activation.retry_state, err):
                         logger.info(
```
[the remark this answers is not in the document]
GabeVillalobos: This is intentional. `report_timeout_errors` controls whether or not we report `ProcessingDeadlineExceeded` errors to Sentry.

PR #617, opened 2026-04-30
feat(taskbroker): Passthrough/raw mode

untitaker:
ref STREAM-882

Introduce passthrough mode, so that a broker can be used to spawn tasks
from any topic with any type of message format. This will make it easier
to migrate existing consumers to be tasks instead, without changing data
layout in prod. For more information refer to the ticket above.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -316,7 +316,10 @@ pub struct Config {
     pub raw_taskname: Option<String>,
 
     /// Processing deadline duration in seconds for raw mode activations.
-    pub raw_processing_deadline_duration: u64,
+    ///
+    /// This is an u16 because 1) we don't want to allow signed numbers 2) it can be cast into i32
+    /// (which we use elsewhere) without error conditions. It doesn't actually have to be that small.
+    pub raw_processing_deadline_duration: u16,
 }
 
 impl Default for Config {
--- a/src/config.rs
+++ b/src/config.rs
@@ -301,6 +301,22 @@ pub struct Config {
 
     /// Maps every application to its worker endpoint, both represented as strings.
     pub worker_map: BTreeMap<String, String>,
+
+    /// Enable passthrough mode for consuming raw bytes from legacy topics.
+    /// In passthrough mode, raw Kafka message bytes are wrapped into TaskActivation.
+    pub passthrough_mode: bool,
+
+    /// The namespace to assign to passthrough activations.
+    pub passthrough_namespace: Option<String>,
+
+    /// The application to assign to passthrough activations.
+    pub passthrough_application: Option<String>,
+
+    /// The taskname to assign to passthrough activations.
+    pub passthrough_taskname: Option<String>,
+
+    /// Processing deadline duration in seconds for passthrough activations.
+    pub passthrough_processing_deadline_duration: u64,
 }
 
 impl Default for Config {
@@ -386,6 +402,11 @@ impl Default for Config {
             callback_addr: "0.0.0.0".into(),
             callback_port: 50051,
             worker_map: [("sentry".into(), "http://127.0.0.1:50052".into())].into(),
+            passthrough_mode: false,
+            passthrough_namespace: None,
+            passthrough_application: None,
+            passthrough_taskname: None,
+            passthrough_processing_deadline_duration: 30,
         }
     }
 }
--- a/src/config.rs
+++ b/src/config.rs
@@ -302,21 +302,21 @@ pub struct Config {
     /// Maps every application to its worker endpoint, both represented as strings.
     pub worker_map: BTreeMap<String, String>,
 
-    /// Enable passthrough mode for consuming raw bytes from legacy topics.
-    /// In passthrough mode, raw Kafka message bytes are wrapped into TaskActivation.
-    pub passthrough_mode: bool,
+    /// Enable raw mode for consuming unstructured Kafka messages.
+    /// In raw mode, Kafka message bytes are wrapped into TaskActivation.
+    pub raw_mode: bool,
 
-    /// The namespace to assign to passthrough activations.
-    pub passthrough_namespace: Option<String>,
+    /// The namespace to assign to raw mode activations.
+    pub raw_namespace: Option<String>,
 
-    /// The application to assign to passthrough activations.
-    pub passthrough_application: Option<String>,
+    /// The application to assign to raw mode activations.
+    pub raw_application: Option<String>,
 
-    /// The taskname to assign to passthrough activations.
-    pub passthrough_taskname: Option<String>,
+    /// The taskname to assign to raw mode activations.
+    pub raw_taskname: Option<String>,
 
-    /// Processing deadline duration in seconds for passthrough activations.
-    pub passthrough_processing_deadline_duration: u64,
+    /// Processing deadline duration in seconds for raw mode activations.
+    pub raw_processing_deadline_duration: u64,
 }
 
 impl Default for Config {
@@ -402,11 +402,11 @@ impl Default for Config {
             callback_addr: "0.0.0.0".into(),
             callback_port: 50051,
             worker_map: [("sentry".into(), "http://127.0.0.1:50052".into())].into(),
-            passthrough_mode: false,
-            passthrough_namespace: None,
-            passthrough_application: None,
-            passthrough_taskname: None,
-            passthrough_processing_deadline_duration: 30,
+            raw_mode: false,
+            raw_namespace: None,
+            raw_application: None,
+            raw_taskname: None,
+            raw_processing_deadline_duration: 30,
         }
     }
 }
--- a/src/kafka/deserialize_raw.rs
+++ b/src/kafka/deserialize_raw.rs
@@ -294,7 +294,13 @@ mod tests {
         let inflight = result.unwrap();
         let activation = TaskActivation::decode(inflight.activation.as_slice()).unwrap();
 
-        assert_eq!(activation.headers.get("trace-id"), Some(&"abc123".to_string()));
-        assert_eq!(activation.headers.get("request-id"), Some(&"req-456".to_string()));
+        assert_eq!(
+            activation.headers.get("trace-id"),
+            Some(&"abc123".to_string())
+        );
+        assert_eq!(
+            activation.headers.get("request-id"),
+            Some(&"req-456".to_string())
+        );
     }
 }
--- a/src/kafka/deserialize_raw.rs
+++ b/src/kafka/deserialize_raw.rs
@@ -77,27 +77,28 @@ fn extract_headers(msg: &OwnedMessage) -> HashMap<String, String> {
 }
 
 /// Encode raw bytes into msgpack format: {"args": [raw_bytes], "kwargs": {}}
-fn encode_raw_params(raw_bytes: &[u8]) -> Result<Vec<u8>, Error> {
+fn encode_raw_params(raw_bytes: &[u8]) -> Vec<u8> {
     let params = RawParams {
         args: (serde_bytes::Bytes::new(raw_bytes),),
         kwargs: HashMap::new(),
     };
 
-    // TODO: send to DLQ instead of skipping
-    rmp_serde::to_vec_named(&params).map_err(|e| anyhow!("Failed to encode msgpack: {}", e))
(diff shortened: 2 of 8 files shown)


src/config.rs
```diff
@@ -301,6 +301,22 @@ pub struct Config {
 
     /// Maps every application to its worker endpoint, both represented as strings.
     pub worker_map: BTreeMap<String, String>,
+
+    /// Enable passthrough mode for consuming raw bytes from legacy topics.
```
evanh:
```suggestion
    /// Enable passthrough mode for consuming raw bytes from raw topics.
```

evanh (approved):

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,227 @@
...
+
+/// Encode raw bytes into msgpack format: {"args": [raw_bytes], "kwargs": {}}
+fn encode_passthrough_params(raw_bytes: &[u8]) -> Result<Vec<u8>, Error> {
+    use serde::Serialize;
+
+    #[derive(Serialize)]
+    struct Params<'a> {
+        args: (&'a [u8],),
+        kwargs: HashMap<(), ()>,
+    }
```
markstory: For my own understanding, why define this struct here vs outside of the function?

src/config.rs
```diff
@@ -301,6 +301,22 @@ pub struct Config {
...
+    /// In passthrough mode, raw Kafka message bytes are wrapped into TaskActivation.
+    pub passthrough_mode: bool,
```
fpacifici: Maybe a nit, but the naming feel backwards. Passthrough means letting things through:
- When consuming a tasks topic we expect activations and do not do anything to them till they are stored. This seems the passthrough mode to me.
- When consuming raw topics we need to pre-process them, into Activations before passing them to the rest of the pipeline. 

Am I getting this wrong ?

src/config.rs
```diff
@@ -301,6 +301,22 @@ pub struct Config {
...
+    pub passthrough_namespace: Option<String>,
+
+    /// The application to assign to passthrough activations.
+    pub passthrough_application: Option<String>,
+
+    /// The taskname to assign to passthrough activations.
+    pub passthrough_taskname: Option<String>,
+
+    /// Processing deadline duration in seconds for passthrough activations.
+    pub passthrough_processing_deadline_duration: u64,
```
fpacifici: What would happen if we did not assign them a deadline? Today there is no explicit deadline apart for max poll time in kafka.
I am not saying we should set one, I'd much rather having a deadline that is explicit and we control rather than letting the infrastructure kill tasks in unexpected way. 
We would have to verify if we have tasks that routinely take long period of times in multi processing or in any other setup that does not cause issues with the max poll time.

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,227 @@
...
+
+    let params = Params {
+        args: (raw_bytes,),
+        kwargs: HashMap::new(),
+    };
+
+    rmp_serde::to_vec_named(&params).map_err(|e| anyhow!("Failed to encode msgpack: {}", e))
```
fpacifici: What is a real world scenario where this error would happen? Would we consider the message invalid and DLQ? Please add a TODO to make this clear if that is the case.

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,227 @@
...
+}
+
+/// Create a deserializer closure for passthrough mode.
+/// Wraps raw Kafka message bytes into a TaskActivation with msgpack-encoded parameters_bytes.
+pub fn new(
+    config: PassthroughConfig,
+) -> impl Fn(Arc<OwnedMessage>) -> Result<InflightActivation, Error> {
+    move |msg: Arc<OwnedMessage>| {
+        let Some(payload) = msg.payload() else {
+            return Err(anyhow!("Message has no payload"));
```
fpacifici: I think we need to be careful before erroring here
https://github.com/getsentry/arroyo/blob/main/arroyo/backends/kafka/consumer.py#L521

Arroyo manages this turning the Null message into an empty binary string.
I doubt anybody process those messages but I cannot verify either.

Would this be a DLQ scenario as well? If not, what is the value in crashing here ?

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,227 @@
...
+        };
+
+        let id = Uuid::new_v4().to_string();
+        let parameters_bytes = encode_passthrough_params(payload)?;
+        let now = Utc::now();
+        let received_at = prost_types::Timestamp {
```
fpacifici, line 108: Do we not have a task timestamp in taskbroker? 
We have consumers that do something with old messages (where the broker timestamp is old), it is convenient, if we only provide the timestamp at which the task is received by the consumer that information (old message) is lost.  Have you considered using the message timestamp here ?

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        let activation_bytes = activation.encode_to_vec();
+        let bucket = bucket_from_id(&id);
+
+        metrics::histogram!(
+            "consumer.passthrough.payload_size_bytes",
+            "namespace" => config.namespace.clone(),
+            "taskname" => config.taskname.clone()
+        )
+        .record(payload.len() as f64);
```
fpacifici, line 137: Are these metrics batched or sampled ? Producing metrics per message is a performance bottleneck

src/config.rs
untitaker: we're passing through.. the raw bytes from the topic. or i guess as per your reading the activations. in the end I agree this is not a good name. "raw mode"? this would be consistent with evan's suggestion to rename "legacy topic" to "raw topic"

src/kafka/deserialize_raw.rs
untitaker: the metrics in taskbroker work different than in arroyo. we already have other metrics that are per-message. the metrics backend aggregates them transparently  and we seem to be fine with that.

(i've always wanted the same thing in arroyo, making metrics collection fundamentally more performant rather than trying to pollute all the instrumentation)

src/kafka/deserialize_raw.rs
untitaker: ah you're right, we should use the produce ts

src/kafka/deserialize_raw.rs
untitaker: the way this error is currently handled there will only be a error log but the consumer doesn't crash. this is consistent with how we handle deserialization in the normal mode too.

there is no dlq yet, I think we would want to use it once it's there, for both modes (i.e. also if an activation fails to parse from the task-topic)

src/kafka/deserialize_raw.rs
untitaker: see next comment about dlq, added the todo

src/config.rs
untitaker: processing deadline is strictly required by the task activation and the entire architecture. it's defaulting to 10 seconds across all tasks, but i felt that that was too short. 30 seconds is more commonly set in task definitions.

i agree that the application should define it and not the infra. however, like with the max-retries (which will be added later), there is currently no way to communicate task metadata back from the worker to the broker. I would like to postpone this work so that it doesn't interfere with other taskbroker rollouts.

src/kafka/deserialize_raw.rs
untitaker: since there's only one place where I use it. this is purely stilistic and somewhat unpopular, i'll move it out.

src/kafka/deserialize_raw.rs
markstory: > there is no dlq yet, I think we would want to use it once it's there, for both modes (i.e. also if an activation fails to parse from the task-topic)

All the current taskbroker deployments do have DLQ topics setup, and taskbroker has a few DLQ producers already (but they aren't in use yet).

src/kafka/deserialize_raw.rs
markstory: Yeah the message produce time could be used as the activation's `received_at`

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        let activation = TaskActivation {
+            id: id.clone(),
+            application: Some(config.application.clone()),
+            namespace: config.namespace.clone(),
+            taskname: config.taskname.clone(),
+            #[allow(deprecated)]
+            parameters: String::new(),
+            parameters_bytes,
+            headers: HashMap::new(),
```
markstory: Do we want to copy any of the original message's headers?

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        Ok(InflightActivation {
+            id,
+            activation: activation_bytes,
+            status: InflightActivationStatus::Pending,
+            partition: msg.partition(),
+            offset: msg.offset(),
+            added_at: now,
+            received_at: received_at_time,
+            processing_deadline: None,
```
markstory, line 147: Should this be `config.processing_deadline_duration`?

src/kafka/deserialize_raw.rs
untitaker: we can forward them in their entirety for now, probably useful. fixed. if we later realize it's a bad idea we can revert.

src/kafka/deserialize_raw.rs
untitaker: `processing_deadline` is set to `None` in the other deserializer too. I'm not sure how it's used, but it's a timestamp. seems like we set it in the db a bunch.

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,300 @@
...
+            claim_expires_at: None,
+            processing_deadline_duration: config.processing_deadline_duration as i32,
```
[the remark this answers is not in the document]
untitaker: it is what it is. this cast exists in both `deserialize_activation.rs` and here, so I think it's fine...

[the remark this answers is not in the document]
untitaker: actually nvm we have control over the config

src/kafka/deserialize_raw.rs
untitaker: Ah it's called "deadletter" in upkeep.

I'll change this to convert to emptystring, like arroyo. This way the failure is pushed downstream into the task, which effectively makes it DLQable.

src/kafka/deserialize_raw.rs
untitaker: mark pointed out we have a dlq, and i took another look at this. I don't think this can happen at all, but if it does I think it's better to crash taskbroker entirely since it's a bug in taskbroker. Updated to panic.

src/kafka/deserialize_raw.rs
```diff
@@ -0,0 +1,326 @@
...
+
+/// Encode raw bytes into msgpack format: {"args": [raw_bytes], "kwargs": {}}
+fn encode_raw_params(raw_bytes: &[u8]) -> Vec<u8> {
+    let params = RawParams {
+        args: (serde_bytes::Bytes::new(raw_bytes),),
+        kwargs: HashMap::new(),
+    };
+
+    // The only condition where this would fail should be a bug in msgpack or taskbroker. In this
+    // case it's better to stop the world instead of trying to recover externally.
```
evanh, line 87: Why is it better to stop the world?

src/kafka/deserialize_raw.rs
evanh: Yeah this is updated when the row is fetched/pushed, based on `processing_deadline_duration`. It's the actual timestamp at which an activation is past the deadline. https://github.com/getsentry/taskbroker/blob/main/src/store/adapters/postgres.rs#L463

src/kafka/deserialize_raw.rs
untitaker: if we just skip the message, we will have to recover it. we can't dlq easily from this place, and I don't think it's worth it to introduce dlqing here, since a failure from msgpack is extremely unlikely to be related to the specific message at all

src/kafka/deserialize_raw.rs
markstory: My bad, I got the two names mixed up.

markstory (approved):

PR #618, opened 2026-04-30
feat(taskbroker): Batch Status Updates

george-sentry:
## Linear
Completes [STREAM-918](https://linear.app/getsentry/issue/STREAM-918/batch-status-updates)

## Description
On the usual workload of 100 millisecond tasks, with the new "claimed" status, we can do around 5K tasks per second in the sandbox. By batching status updates, we reduce DB load, making all queries take less time. This can increase throughput by 1K to 2K tasks per second.


george-sentry:
Since we may want to treat claimed → processing updates the same way, I'm actually going to create a more general `Flusher` struct that can be used by both push threads and the gRPC server.

src/flusher.rs
```diff
@@ -0,0 +1,62 @@
...
+use std::pin::Pin;
+use std::time::Duration;
+
+use anyhow::Result;
+use tokio::sync::mpsc::Receiver;
+
+/// Run flusher that receives values of type T from a channel and flushes
+/// them using the provided async `flush` function either when the batch is
+/// full or when the max flush interval has elapsed.
+pub async fn run_flusher<T, F>(
```
george-sentry, line 12: I created this function because I'm also planning to batch claimed → processing updates in the push pool, which will use basically identical machinery.

src/store/adapters/sqlite.rs
```diff
@@ -733,6 +760,32 @@ impl InflightActivationStore for SqliteActivationStore {
...
+
+        let sql = format!(
+            "DELETE FROM inflight_taskactivations WHERE id IN ({})",
+            placeholders.join(", ")
+        );
+
+        let mut q = sqlx::query(&sql);
+        for id in ids {
+            q = q.bind(id);
+        }
```
markstory: Wouldn't using the `builder.separated()` let you iterate the id list only once and make fewer intermediary string values?

src/store/adapters/sqlite.rs
```diff
@@ -706,6 +706,33 @@ impl InflightActivationStore for SqliteActivationStore {
...
+    ) -> Result<(), Error> {
+        if ids.is_empty() {
+            return Ok(());
+        }
+
+        let mut conn = self.acquire_write_conn_metric("set_status_batch").await?;
+
+        let placeholders: Vec<String> = (0..ids.len()).map(|i| format!("?{}", i + 2)).collect();
+        let sql = format!(
+            "UPDATE inflight_taskactivations SET status = ?1 WHERE id IN ({})",
```
markstory: Should these placeholders be `$1` style?

src/store/adapters/postgres.rs
```diff
@@ -664,6 +685,28 @@ impl InflightActivationStore for PostgresActivationStore {
...
+        if ids.is_empty() {
+            return Ok(0);
+        }
+
+        let mut conn = self
+            .acquire_write_conn_metric("delete_activations_by_id")
+            .await?;
+
+        let mut query_builder =
+            QueryBuilder::new("DELETE FROM inflight_taskactivations WHERE id = ANY(");
```
markstory: Why use `ANY()` here, while we use `IN (...)` in most other places?

src/store/adapters/postgres.rs
```diff
@@ -637,6 +637,27 @@ impl InflightActivationStore for PostgresActivationStore {
...
+        ids: &[String],
+        status: InflightActivationStatus,
+    ) -> Result<(), Error> {
+        if ids.is_empty() {
+            return Ok(());
+        }
+
+        let mut conn = self.acquire_write_conn_metric("set_status_batch").await?;
+
+        sqlx::query("UPDATE inflight_taskactivations SET status = $1 WHERE id = ANY($2)")
```
markstory: Why `ANY()` over `IN`?

src/grpc/status_flusher.rs
```diff
@@ -0,0 +1,136 @@
...
+                            flush_buffer(&store, &mut buffer).await;
+                        }
+
+                        debug!("Status flusher shutting down...");
+                        break;
+                    }
+                }
+            }
+
+            _ = interval.tick() => {
```
fpacifici: When does this trigger ?

src/grpc/status_flusher.rs
```diff
@@ -0,0 +1,136 @@
...
+                ?status,
+                ?count,
+                error = ?e,
+                "Failed to flush status batch"
+            );
+
+            // Push failed updates back into the buffer so they can be retried on next flush
+            for id in ids {
+                buffer.push((id, status));
+            }
```
fpacifici: Let's say there is a DB issue, would we keep appending to the buffer indefinitely? I think we should add a limit after which we stop and retry on the DB.

src/grpc/server.rs
```diff
@@ -194,3 +208,58 @@ impl ConsumerService for TaskbrokerServer {
         res
     }
 }
+
+pub type StatusUpdate = (String, InflightActivationStatus);
+
+pub async fn flush_status_updates(
```
evanh: I don't think this function should be in the server file. This feels like maybe it should be part of the store itself.

src/grpc/server_tests.rs
```diff
@@ -21,11 +21,17 @@ async fn test_get_task_push_mode_returns_permission_denied() {
         ..Config::default()
     });
 
-    let service = TaskbrokerServer { store, config };
+    let service = TaskbrokerServer {
```
evanh, line 25: Not reviewing the whole file, are there some tests running that actually use `status_tx` as not None?

src/flusher.rs
```diff
@@ -0,0 +1,62 @@
...
+    let mut buffer: Vec<T> = Vec::with_capacity(batch_size);
+
+    loop {
+        tokio::select! {
+            msg = rx.recv() => {
+                match msg {
+                    Some(v) => {
+                        buffer.push(v);
+
+                        while let Ok(update) = rx.try_recv() {
```
evanh: Why do this again here? Won't this update get processed on the next loop?

src/main.rs
```diff
@@ -192,10 +192,33 @@ async fn main() -> Result<(), Error> {
...
+    let (status_tx, status_flush_task) = if config.delivery_mode == DeliveryMode::Push {
+        let (tx, rx) = tokio::sync::mpsc::channel(config.status_flush_batch_size);
+
+        let flusher_store = store.clone();
+        let flusher_config = config.clone();
+
+        let handle = tokio::spawn(async move {
+            flusher::run_flusher(
+                rx,
+                flusher_config.status_flush_batch_size,
```
evanh: I'm not sure this is right. Won't this create another listener on the channel? Does that mean we will get duplicates?

When we are shutting down presumably we won't get a full batch. That means we are waiting at least `status_flush_internal_ms` before the application can shut down safely. I would be inclined to have this value be much lower when we are shutting down to try and clear the batch as quickly as possible.

src/store/adapters/sqlite.rs
george-sentry: Yeah, either that or it should've used `push_bind` to avoid placeholders altogether. Corrected.

src/store/adapters/sqlite.rs
george-sentry: Yepp, did that instead.

src/store/adapters/postgres.rs
george-sentry: No big reason. Just a good way to avoid binding every item in the list separately, so cleaner code I guess?

src/store/adapters/postgres.rs
george-sentry: This way, I can avoid binding every item in the list separately, so cleaner code I guess?

src/flusher.rs
george-sentry: Yes, but in theory, it'll be slower if we wait for the next loop iteration, since it'll need to be reawakened. The idea is, while we're already awake, we might as well empty the whole channel rather than going back to sleep just to be awakened a moment later for the next item in the channel.

george-sentry: However, I think there was a bug here that could result in this loop spinning for a very long time. I fixed it by adding a `while buffer.len() < batch_size` condition.

src/grpc/server.rs
```diff
@@ -194,3 +208,58 @@ impl ConsumerService for TaskbrokerServer {
...
+                    "Failed to flush status batch"
+                );
+
+                // Push failed updates back into the buffer so they can be retried on next flush
+                for id in ids {
+                    buffer.push((id, status));
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
george-sentry: No longer in scope of this PR.

src/grpc/status_flusher.rs
george-sentry: This was actually dead code, but similar logic now lives elsewhere.

No, we only append to the buffer while it hasn't reached the desired batch size. So if there's a DB issue, here's what should happen.

1. Timer runs out or buffer fills up → call `flush` (this function)
2. As long as `flush` is running, the (now empty) buffer does not receive any more IDs
3. Flush fails because store is unresponsive or some other problem
4. IDs are pushed back onto the buffer (which was emptied right before attempting the flush)
5. Flush function exits

So if the DB has a problem, we will keep retrying the same batch of IDs over and over again until it succeeds.

src/grpc/status_flusher.rs
george-sentry: This code now lives in `flusher.rs`, but it contains a similar loop. This condition triggers every `interval_ms` after the previous tick.

The flusher only handles the tick when `select!` actually chooses this branch. If messages keep arriving and the `rx.recv()` arm keeps winning before the tick is ready, the tick still advances in the background. When the tick is ready and this arm is selected, the buffer is flushed.

src/store/adapters/sqlite.rs
```diff
@@ -706,6 +706,33 @@ impl InflightActivationStore for SqliteActivationStore {
...
+            placeholders.join(", ")
+        );
+
+        let mut q = sqlx::query(&sql).bind(status);
+        for id in ids {
+            q = q.bind(id);
```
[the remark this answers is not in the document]
george-sentry: Outdated.

src/grpc/server.rs
```diff
@@ -194,3 +208,58 @@ impl ConsumerService for TaskbrokerServer {
...
+
+    metrics::gauge!(format!("grpc_server.flush_status_updates.success")).set(success as f64);
+    metrics::gauge!(format!("grpc_server.flush_status_updates.fail")).set(fail as f64);
+}
```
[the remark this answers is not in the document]
george-sentry: No longer in scope for this PR.

src/grpc/server.rs
george-sentry: Sure. In its own file within the store module, or within an existing file?

src/main.rs
george-sentry: No, `rx` here is the only listener on this channel as it's an MPSC channel. And in this case, there is also only one writer - the gRPC server.

src/main.rs
```diff
@@ -278,6 +302,10 @@ async fn main() -> Result<(), Error> {
         departure = departure.on_completion(log_task_completion("fetch_task", task));
     }
 
+    if let Some(task) = status_update_task {
+        departure = departure.on_completion(log_task_completion("status_update_task", task));
+    }
```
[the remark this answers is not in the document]
markstory: Seems like a relevant problem that should be addressed.

src/grpc/server.rs
```diff
@@ -97,10 +101,20 @@ impl ConsumerService for TaskbrokerServer {
...
         }
+
         if status == InflightActivationStatus::Failure {
             metrics::counter!("grpc_server.set_status.failure").increment(1);
         }
 
+        if let Some(ref tx) = self.update_tx {
+            tx.send((id, status))
+                .await
+                .map_err(|_| Status::internal("Status update channel closed"))?;
```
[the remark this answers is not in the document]
george-sentry: Pull mode workers aren't supposed to be using this path.

src/fetch/tests.rs
```diff
@@ -122,6 +122,14 @@ impl InflightActivationStore for MockStore {
         unimplemented!()
     }
 
+    async fn set_status_batch(
+        &self,
+        _ids: &[String],
+        _status: InflightActivationStatus,
+    ) -> Result<u64, Error> {
+        Ok(0)
```
untitaker: should this be `unimplemented!()` liek the other functions?

src/grpc/server.rs
```diff
@@ -194,3 +208,81 @@ impl ConsumerService for TaskbrokerServer {
...
+    if buffer.is_empty() {
+        return;
+    }
+
+    let updates = std::mem::take(buffer);
+    let mut by_status: HashMap<InflightActivationStatus, Vec<String>> = HashMap::new();
+
+    for (id, status) in updates {
+        by_status.entry(status).or_default().push(id);
+    }
```
untitaker:
```suggestion
    let mut by_status: HashMap<InflightActivationStatus, Vec<String>> = HashMap::new();

    for (id, status) in buffer.drain(..) {
        by_status.entry(status).or_default().push(id);
    }
```

src/fetch/tests.rs
george-sentry: Yes, fixed. Thanks!

src/grpc/server.rs
```diff
@@ -194,3 +208,80 @@ impl ConsumerService for TaskbrokerServer {
...
+                    ?status,
+                    requested,
+                    error = ?e,
+                    "Failed to flush status batch from server"
+                );
+
+                // Push failed updates back into the buffer so they can be retried on next flush
+                for id in ids {
+                    buffer.push((id, status));
+                }
```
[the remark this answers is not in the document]
george-sentry: I don't think this is true. Yes, it's unbounded, but we're literally just rebuffering the same tasks that were already there a moment ago. And while this function is running, no more tasks are added to the queue.

src/flusher.rs
```diff
@@ -0,0 +1,62 @@
+use std::future::Future;
+use std::pin::Pin;
+use std::time::Duration;
+
+use anyhow::Result;
+use tokio::sync::mpsc::Receiver;
+
+/// Run flusher that receives values of type T from a channel and flushes
+/// them using the provided async `flush` function either when the batch is
+/// full or when the max flush interval has elapsed.
```
markstory: Would be good to document that this wrapper doesn't handle draining the buffer, and that the flush function is responsible for that.

markstory (commented):
Makes sense to me. I didn't see any paths that would lead to data loss. The additional queue could result in writes that were accepted from the client perspective to be lost during a crash. While we'll run the task an additional time, we shouldn't lose any data, as the activations will still be in postgres.

markstory (approved):

src/fetch/mod.rs
```diff
@@ -122,10 +122,11 @@ impl<T: TaskPusher + Send + Sync + 'static> FetchPool<T> {
                         }
 
                         _ = async {
+                            let start = Instant::now();
```
fpacifici, line 125: Why this change ?

george-sentry:
> The additional queue could result in writes that were accepted from the client perspective to be lost during a crash. While we'll run the task an additional time, we shouldn't lose any data, as the activations will still be in Postgres.

True. I also thought about it this way.

If I batch updates, `set_task_status` will run faster so the taskworker's internal result queue will drain quicker. If the taskbroker crashes, the queued updates will be lost.

If I don't batch updates, `set_task_status` will run slower so the taskworker's internal result queue will drain slower. If the taskworker crashes, the queued updates will be lost.

Regardless, the outcome is roughly the same -- a certain number tasks are lost and executed a second time. The only difference is whether they are queued in the taskbroker or the taskworker.

src/flusher.rs
george-sentry: Done!

src/grpc/server.rs
fpacifici, line 283: Let's say the request fails because the DB has a problem (not a transient one) or it is saturated. 
Are we going to pile up requests till the broker goes out of memory? This would leave the database in an inconsistent state that will be very hard to troubleshoot. 

Do we go through this code only when the worker updates the status or also when we claim the task or reset them during the upkeep ? Upkeep and claim phase have an easier and more deterministic way to batch requests, so they should not go through this.

src/fetch/mod.rs
george-sentry: Not sure. I'm moving it back to where it was in #637.

src/grpc/server.rs
george-sentry: I don't think that'll happen because the buffer is bounded, meaning requests won't pile up. It's not obvious here because the logic that limits buffer size is in `flusher.rs`.

When I simulated various database issues during testing (such as high latency and connection reset errors), this did not appear to be a problem. Taskbroker was able to recover on its own as soon as those issues were resolved.

george-sentry: But it may be worth testing again because that was a while ago.

george-sentry: Correction... there was a bug in the buffering logic in `flusher.rs` that would result in unbounded growth. I'm fixing it in #637.

PR #621, opened 2026-05-04
ref: Change PushPool to use a worker factory for easier test mocking

evanh:
This change allows the worker to be easily mocked in the pool. This allows writing of more comprehensive tests (which are also added in this PR).

### DIFF
--- a/src/push/mod.rs
+++ b/src/push/mod.rs
@@ -1,5 +1,7 @@
 use chrono::Utc;
 use std::collections::HashMap;
+use std::future::Future;
+use std::pin::Pin;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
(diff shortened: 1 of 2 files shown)


src/push/mod.rs
```diff
@@ -23,6 +26,12 @@ use crate::store::traits::InflightActivationStore;
 
 type HmacSha256 = Hmac<Sha256>;
 
+type WorkerFactory = Arc<
+    dyn Fn(String) -> Pin<Box<dyn Future<Output = Result<Box<dyn WorkerClient + Send>>> + Send>>
```
untitaker, line 30: Instead of `Arc<dyn Fn>` I believe this can be `fn`, because the closure you're passing doesn't capture variables from the outside. this would remove one layer of allocations

untitaker (approved):

src/push/mod.rs
evanh: There are some tests that do capture outside variables. I'm going to leave this as is for now, I spent an hour trying to untangle the tests and I don't think it's worth more time.

PR #623, opened 2026-05-07
feat(taskbroker-client): Add pass_headers option to task registration

untitaker:
## Summary

- Adds `pass_headers` parameter to `@namespace.register()` decorator
- When `pass_headers=True`, the task function receives activation headers as a `headers` keyword argument (`dict[str, str]`)
- Provides a simpler alternative to implementing a full `ContextHook` for tasks that just need direct access to header values

We need this kind of functionality when using raw mode together with the profiles consumer. Mainly for killswitches.

ref STREAM-882

## Usage

```python
from taskbroker_client.types import TaskHeaders

@namespace.register(name="my.task", pass_headers=True)
def my_task(org_id: int, headers: TaskHeaders) -> None:
    tenant_id = headers.get("x-tenant-id", "")
    trace_id = headers.get("sentry-trace", "")
```

## Test plan

- [x] Added unit tests for `pass_headers` attribute on Task
- [x] Added tests for ALWAYS_EAGER mode (empty headers dict passed)
- [x] Added worker integration test verifying headers are passed during execution
- [x] All 158 tests pass

### DIFF
--- a/clients/python/src/taskbroker_client/registry.py
+++ b/clients/python/src/taskbroker_client/registry.py
@@ -90,6 +90,7 @@ class TaskNamespace:
         compression_type: CompressionType = CompressionType.PLAINTEXT,
         report_timeout_errors: bool = True,
         silenced_exceptions: tuple[type[BaseException], ...] | None = None,
+        pass_headers: bool = False,
     ) -> Callable[[Callable[P, R]], Task[P, R]]:
         """
         Register a task.
@@ -121,6 +122,9 @@ class TaskNamespace:
             Enable reporting of ProcessingDeadlineExceededError to Sentry.
         silenced_exceptions: tuple[type[BaseException], ...] | None
             A tuple of exception types that will not be reported by Sentry.
+        pass_headers: bool
+            If True, the task function will receive task activation headers
+            as a keyword argument named `headers` (dict[str, str]).
         """
 
         def wrapped(func: Callable[P, R]) -> Task[P, R]:
@@ -141,6 +145,7 @@ class TaskNamespace:
                 compression_type=compression_type,
                 report_timeout_errors=report_timeout_errors,
                 silenced_exceptions=silenced_exceptions,
+                pass_headers=pass_headers,
             )
             # TODO(taskworker) tasks should be registered into the registry
             # so that we can ensure task names are globally unique
@@ -234,6 +239,7 @@ class ExternalNamespace(TaskNamespace):
         compression_type: CompressionType = CompressionType.PLAINTEXT,
         report_timeout_errors: bool = True,
         silenced_exceptions: tuple[type[BaseException], ...] | None = None,
+        pass_headers: bool = False,
     ) -> Callable[[Callable[P, R]], ExternalTask[P, R]]:
         """
         Register an external task stub.
@@ -285,6 +291,7 @@ class ExternalNamespace(TaskNamespace):
                 compression_type=compression_type,
                 report_timeout_errors=report_timeout_errors,
                 silenced_exceptions=silenced_exceptions,
+                pass_headers=pass_headers,
             )
             self._registered_tasks[name] = task
             return task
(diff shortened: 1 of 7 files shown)


clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -402,7 +402,10 @@ def _execute_activation(
                         ):
                             for hook in context_hooks:
                                 stack.enter_context(hook.on_execute(headers))
-                        task_func(*args, **kwargs)
+                        if task_func.pass_headers:
+                            task_func(*args, headers=headers, **kwargs)
```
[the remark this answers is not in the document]
evanh: I don't think we should silently overwrite the user's arguments. This should raise an Exception.

clients/python/src/taskbroker_client/registry.py
```diff
@@ -141,6 +145,7 @@ def wrapped(func: Callable[P, R]) -> Task[P, R]:
                 compression_type=compression_type,
                 report_timeout_errors=report_timeout_errors,
                 silenced_exceptions=silenced_exceptions,
+                pass_headers=pass_headers,
```
markstory, line 148: Since we're going to inject parameters into the task, should we validate that the task doesn't already have a `headers` parameter with an incompatible type?

markstory (commented):
I think the general design is reasonable. Another potential solution would be to pass the entire activation into the task, and let userland code pick the headers out. I prefer passing just the headers over the the entire activation though.

clients/python/src/taskbroker_client/registry.py
untitaker: > Another potential solution would be to pass the entire activation into the task, 

as in, you have a `pass_activation` option?

> Since we're going to inject parameters into the task, should we validate that the task doesn't already have a headers parameter with an incompatible type?

this feels magical but i can do it

untitaker: I think we can enforce that the headers kwarg is there explicitly. `def mytask(**kwargs):` is then just invalid.

markstory: > I think we can enforce that the headers kwarg is there explicitly. def mytask(**kwargs): is then just invalid.

I was more thinking of the scenario where a task has an incompatible `headers` parameter like:

```python
def some_task(headers: set[str]):
```

markstory: > as in, you have a pass_activation option?

Yeah, but as I think about it more I like this idea less, and `pass_headers` is the better solution.

untitaker: i think the current version covers this. it requires the header arg to be there, to have the right type, and to be explicit

clients/python/tests/test_task.py
```diff
@@ -465,3 +465,79 @@ def multi_task() -> None:
     activation = multi_task.create_activation([], {})
     assert activation.headers["x-test-context"] == "dispatched"
     assert activation.headers["x-another"] == "also-here"
+
```
evanh, line 468: Can you add a test for the case when the user passes `headers` as one of the `kwargs` to their function?

evanh (approved):

markstory (approved):
Looks good to me. The test addition that Evan mentioned would be a good one to have.

clients/python/tests/test_task.py
untitaker: ok. testing for a specific error doesn't make a lot of sense if there's no explicit errorhandling for it, so I added more checks to the code too. i think it's a bit bloated now, but probably more user-friendly.

clients/python/src/taskbroker_client/task.py
```diff
@@ -54,6 +55,42 @@ def _get_parameters_format() -> ParametersFormat:
...
+        )
+
+    if param.annotation is not inspect.Parameter.empty:
+        origin = get_origin(param.annotation)
+        if origin is None:
+            origin = param.annotation
+        if origin not in expected_types:
+            raise TypeError(
+                f"{context}: {param_name!r} parameter has type {param.annotation!r}. "
+                f"Expected one of: {', '.join(t.__name__ for t in expected_types)}."
```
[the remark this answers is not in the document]
untitaker: I don't care. Pick better types.

PR #627, opened 2026-05-08
fix(iswf): Reverts change to reporting logic that silences NoRetriesRemaining exceptions

GabeVillalobos:
This reporting is valuable for other teams to triage issues with their tasks and retries, even through silenced exceptions. Knowing when a task has exceeded its maximum number of retries means that the task has effectively transitioned to a terminal state and will not be reattempted if no deadletter is set. While this will increase a bit of noise, it'll provide better ideas of failure rates in our product.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/workerchild.py
+++ b/clients/python/src/taskbroker_client/worker/workerchild.py
@@ -281,18 +281,17 @@ def child_process(
                         next_state = TASK_ACTIVATION_STATUS_RETRY
                     elif retry.max_attempts_reached(inflight.activation.retry_state):
                         with sentry_sdk.isolation_scope() as scope:
-                            if should_capture_error:
-                                retry_error = NoRetriesRemainingError(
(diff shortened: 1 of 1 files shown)


clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -281,18 +281,17 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
-                                scope.fingerprint = [
-                                    "taskworker.no_retries_remaining",
-                                    inflight.activation.namespace,
-                                    inflight.activation.taskname,
-                                ]
-                                scope.set_transaction_name(inflight.activation.taskname)
-                                sentry_sdk.capture_exception(retry_error)
+                            retry_error = NoRetriesRemainingError(
+                                f"{inflight.activation.taskname} has consumed all of its retries"
+                            )
```
markstory, line 286: So even for a silenced error we would capture an error when the task has run out of retries and will be discarded :+1:

kcons (approved):

PR #630, opened 2026-05-11
feat(taskbroker): Implement retry support for raw topics

untitaker:
## Summary

Implement two independent features in this PR:

* `kafka_retry_topic` so that one can send activations from retried tasks into a separate topic. That topic never contains raw messages.
* Ability to get `max_retries` from the worker. Previously the producer would configure `max_retries` and send them as part of the activation, but that cannot work with raw topics, so an architecture change is needed.

See [Architecture doc](https://www.notion.so/3448b10e4b5d80e7a1efee6145d504c2) → Stage 4 for full context.

## Dependencies

**Depends on**: https://github.com/getsentry/sentry-protos/pull/251 (adds `max_retries` field to `SetTaskStatusRequest`)

ref STREAM-981


src/grpc/server.rs
```diff
@@ -101,6 +101,16 @@ impl ConsumerService for TaskbrokerServer {
             metrics::counter!("grpc_server.set_status.failure").increment(1);
         }
 
+        // If status is Retry and max_retries is provided, update the activation's retry_state.
+        // This allows workers to communicate retry policy for tasks from raw topics.
+        if status == InflightActivationStatus::Retry {
+            if let Some(max_retries) = request.get_ref().max_retries {
+                if let Err(e) = self.store.update_retry_state(&id, max_retries).await {
```
george-sentry: How often will this happen? If it happens on every request, throughput will drop significantly. If it happens only occasionally, it should be fine.

untitaker: on every retry, we need to update the task activation. but i think i'll redesign this so that the task activation is updated in one go (max_retries and retry_count are updated using the same update stmt)

src/grpc/server.rs
```diff
@@ -106,7 +106,14 @@ impl ConsumerService for TaskbrokerServer {
             metrics::counter!("grpc_server.set_status.failure").increment(1);
         }
 
-        if let Some(ref tx) = self.update_tx {
+        let max_attempts = request.get_ref().max_attempts;
```
george-sentry, line 109: I already asked in a different comment, but I think that code is now outdated so I'll ask here again. How often will there be a `max_attempts` field on the message? One out of every... 10? 100? 1000? If it's going to be present often, we'll need to rethink how batching works. Because batching is necessary to reach high throughput.

src/store/traits.rs
```diff
@@ -71,11 +71,13 @@ pub trait InflightActivationStore: Send + Sync {
...
     async fn mark_activation_processing(&self, id: &str) -> Result<(), Error>;
 
-    /// Update the status of a specific activation
+    /// Update the status of a specific activation.
+    /// If max_attempts is provided (for Retry status), also updates the activation's retry_state.
     async fn set_status(
         &self,
         id: &str,
         status: InflightActivationStatus,
+        max_attempts: Option<u32>,
```
george-sentry, line 80: Adding this argument here sort of changes what `set_status` does. Before, it only set the status. Now it _also_ updates `retry_state`. Most calls to `set_status` throughout the tests pass `None` because in most cases, it isn't needed. Perhaps there should be a separate method to handle this scenario? Like `set_retry_status` or `set_retriable_status`? Whatever name makes sense.

src/kafka/deserialize.rs
```diff
@@ -12,27 +13,40 @@ use super::deserialize_raw::{self, RawConfig};
 pub struct DeserializeConfig {
     activation_config: DeserializeActivationConfig,
     raw_config: Option<RawConfig>,
+    /// Retry topic always contains activations, even in raw_mode.
```
george-sentry, line 16: What does this comment mean? Why is the raw mode distinction important? I thought raw mode was the _only_ mode in which we used the retry topic?

george-sentry: I think I understand now. You mean that every message in the retry topic is guaranteed to be an activation even in raw mode, whereas messages in the "normal" topic in raw mode may not be?

src/store/adapters/postgres.rs
```diff
@@ -651,6 +655,30 @@ impl InflightActivationStore for PostgresActivationStore {
...
+                sqlx::query("UPDATE inflight_taskactivations SET activation = $1 WHERE id = $2")
+                    .bind(activation.encode_to_vec())
+                    .bind(id)
+                    .execute(&mut *conn)
+                    .await?;
+            }
+        }
+
         Ok(Some(row.into()))
     }
```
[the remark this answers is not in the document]
george-sentry: Possibly valid. I could see that happening if there's a database issue between the queries, or if another component (like upkeep) fiddles around with the tasks in between. What do you think, Markus?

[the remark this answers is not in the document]
george-sentry: Could be another good reason to separate logic for retry tasks into a separate method like `set_retry_status` - see my other comment on that.

src/store/adapters/postgres.rs
```diff
@@ -651,6 +655,30 @@ impl InflightActivationStore for PostgresActivationStore {
...
+                retry_state.max_attempts = max_attempts;
+
+                sqlx::query("UPDATE inflight_taskactivations SET activation = $1 WHERE id = $2")
+                    .bind(activation.encode_to_vec())
+                    .bind(id)
+                    .execute(&mut *conn)
+                    .await?;
+            }
+        }
+
```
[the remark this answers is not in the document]
george-sentry: Sounds valid.

clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -445,6 +445,7 @@ def update_task(
             id=processing_result.task_id,
             status=processing_result.status,
             fetch_next_task=fetch_next_task,
+            max_attempts=processing_result.max_attempts,
```
fpacifici, line 448: Do we have an example of how the task will define this value ?

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -261,6 +262,7 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
                 retry = task_func.retry
                 if retry and retry.should_retry(inflight.activation.retry_state, err):
                     next_state = TASK_ACTIVATION_STATUS_RETRY
+                    max_attempts_val = retry._times + 1
```
fpacifici: Not following this.  Why adding 1 ?

src/grpc/server.rs
fpacifici: This should be present always in tasks that run out of standard topics. I'd go further and say that maybe we should simplify the system by moving all tasks to specify the retries via the set_status method so we do not maintain more than one implementation.

Why does this affect the way batch is implemented ?

src/grpc/server.rs
```diff
@@ -106,7 +106,14 @@ impl ConsumerService for TaskbrokerServer {
...
+
+        // Use batching channel if available and we don't need to update retry state.
+        // If max_attempts is Some, we can't use batching API to update the activation, and have to
+        // fall back to individual set_status.
+        if let Some(ref tx) = self.update_tx
+            && max_attempts.is_none()
+        {
```
fpacifici, line 118: Taking a step back on the retry topic. 
If the retry topic contains activations rather than the original message, why do we even need a topic ? can't we just store the activation in the DB as pending and treat it like a task for the rest of its lifetime ? 

I recognize this is a departure from the original intent of this PR, but  it seems a lot simpler to me to manage it this way. The idea of the topic, to me, was meant to use it as a DLQ as well. Am I missing something ?

fpacifici (commented):
One high level comment 
https://github.com/getsentry/taskbroker/pull/630/changes#r3262500919

src/grpc/server.rs
untitaker: > can't we just store the activation in the DB as pending and treat it like a task for the rest of its lifetime ?

I am missing rationale for why the retry system was set up this way to begin with. Your suggestion also applies to how regular tasks work. I would not want to special-case raw-mode to handle retries fundamentally differently than regular tasks.

My guess is that we wanted to keep the size of a database under control, therefore pruning queued retries out of the DB and putting them back into Kafka. If we say that this is not really a concern with AlloyDB then that's fine, but we'd have to validate that IMO

I can explore this option in another PR, but not sure we should roll it out without having more context from folks who originally worked on taskbroker.

> The idea of the topic, to me, was meant to use it as a DLQ as well

That is yet another topic. It can stay or go away regardless of what we decide wrt retries.

src/grpc/server.rs
untitaker: The field will be there unconditionally on all retry status updates. So retries are effectively not batched. I can implement batching for retries but this would increase complexity.

Note that not every retry status update results in an additional DB query. `max_attempts` is only set in the DB if it wasn't there before.

> This should be present always in tasks that run out of standard topics.

It's optional here for the sake of rollout. We can gradually increase the amount of tasks that send `max_attempts` through the worker and observe its impact on the broker. (the rollout mechanism isn't implemented here)

src/store/traits.rs
untitaker: It used to be a separate method, but in order to allow storage backends to optimize these queries I think we should pass it in a single method call.

Right now the postgres backend runs two queries, but one could imagine altering the internal schema so that setting the status and updating the max_attempts can be done in a single UPDATE. But the implementation already benefits from the fact that everything happens in set_status, because it can skip the second UPDATE for task activations that already have max_retries set.

src/kafka/deserialize.rs
untitaker: > I think I understand now. You mean that every message in the retry topic is guaranteed to be an activation even in raw mode, whereas messages in the "normal" topic in raw mode may not be?

Every message in the retry topic is a TaskActivation protobuf regardless of whether the consumer is in raw mode or normal mode. This is because we need to store the retry count in the topic, somehow.

src/grpc/server.rs
untitaker: @markstory @enochtangg do you remember why we didn't "just" stick retries into the DB and produce them back into kafka?

clients/python/src/taskbroker_client/worker/workerchild.py
markstory: `max_attempts` shouldn't change. If we increase that value, the task will get an increasing number of retries and might not ever exhaust all retries.

src/grpc/server.rs
fpacifici: > Your suggestion also applies to how regular tasks work.

I thought we picked up the task from the database to do the retry. @george-sentry  did we change something in the push model ?

clients/python/src/taskbroker_client/worker/client.py
untitaker: nothing changes about how this value is defined. it's already public API: https://github.com/getsentry/sentry/blob/9ee21b63ae7bcd3bf9a002e077e7fc73b860c656/src/sentry/deletions/tasks/scheduled.py#L102-L103

clients/python/src/taskbroker_client/worker/workerchild.py
untitaker: this is the constant retry.times from the public API: https://github.com/getsentry/sentry/blob/9ee21b63ae7bcd3bf9a002e077e7fc73b860c656/src/sentry/deletions/tasks/scheduled.py#L102-L103

`retry._times` is a constant, so +1 is a constant. we're not continuously incrementing.

src/grpc/server.rs
untitaker: from slack:

> @markstory: Keeping retries in sqlite/postgres could work but those retries will consume slots in sqlite/postgres. It will mean that a retry with no delay runs right away though instead of 'later' when it is found again in the topic.

> @enochtangg: Another benefit I remember was because it resets the latency metric. Since latency is task dispatched - kafka receive latency, re-producing in kafka means we don't need to somehow fix that.

> @fpacifici: the latency metric argument is important. I think we can keep it as it is and use the topic.  It is true that in AlloyDB there will be more room, but we will have the sqlite around for a while. No need to make the change

So I think I won't change anything here.

src/store/adapters/postgres.rs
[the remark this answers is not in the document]
untitaker: If we split up the method entirely, then it'll be actually harder to fix those race conditions.

src/kafka/deserialize.rs
```diff
@@ -12,27 +13,40 @@ use super::deserialize_raw::{self, RawConfig};
...
 
     move |msg: Arc<OwnedMessage>| {
+        // Messages from the retry topic are always activations
+        if let Some(ref retry_topic) = retry_topic
+            && msg.topic() == retry_topic
+        {
+            return activation_deserializer(msg);
+        }
+
+        // For main topic: use raw deserializer in raw_mode, else activation deserializer
```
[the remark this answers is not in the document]
untitaker: ok we can lint against that

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -378,6 +378,14 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
                         host=inflight.host,
                         receive_timestamp=inflight.receive_timestamp,
+                        # Send max_attempts if this is a retry. Don't send it
+                        # on every task as this codepath is relatively
+                        # unoptimized on the broker side.
+                        max_attempts=(
+                            task_func.retry._times + 1
+                            if task_func.retry and next_state == TASK_ACTIVATION_STATUS_RETRY
+                            else None
+                        ),
```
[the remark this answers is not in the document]
untitaker: damn, i thought we had a nonzero default value for this. will probably add `delay_on_retry` to `SetTaskStatusRequest`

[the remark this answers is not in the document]
untitaker: https://github.com/getsentry/sentry-protos/pull/262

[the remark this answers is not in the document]
markstory: Why does the max_attempts go up each time? Shouldn't `max_attempts` draw from the task definition?

src/grpc/server.rs
markstory: We can store retries in the sql db instead of going through the topic again. In the sqlite driver, if the number of retry/delayed tasks gets high (8192 by default) ingestion pauses. We did that to limit the amount of state in sqlite. We could easily tune that higher given how large the disks for sqlite are, and how much capacity we would have in postgres.

I think we could work around Enoch's points on latency if we needed to. 

If we want to switch to retries not going through Kafka, I think we should handle retries consistently for all workloads so that we don't introduce more complexity. Perhaps we can revisit how retries are stored separate of these changes?

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -378,6 +378,19 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
                         host=inflight.host,
                         receive_timestamp=inflight.receive_timestamp,
+                        # Send max_attempts and delay_on_retry if this is a retry.
+                        # Don't send it on every task as this codepath is relatively
+                        # unoptimized on the broker side.
+                        max_attempts=(
+                            task_func.retry._times + 1
+                            if task_func.retry and next_state == TASK_ACTIVATION_STATUS_RETRY
+                            else None
+                        ),
```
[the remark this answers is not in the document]
untitaker: this is incorrect because mx_attempts semantically is max_retries + 1

src/store/traits.rs
markstory: Needing two RPC calls is something we should avoid anyways.

markstory (approved):

clients/python/src/taskbroker_client/worker/workerchild.py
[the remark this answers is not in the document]
untitaker: > Why does the max_attempts go up each time? Shouldn't max_attempts draw from the task definition?

it doesn't go up. `task_func.retry._times` is what is declared by the user in the task decorator. +1 to also count the first attempt.

src/main.rs
```diff
@@ -158,11 +168,19 @@ async fn main() -> Result<(), Error> {
         let consumer_store = store.clone();
         let consumer_config = config.clone();
         let runtime_config_manager = runtime_config_manager.clone();
+
+        // Build list of topics to consume from
+        let mut topics_to_consume = vec![consumer_config.kafka_topic.clone()];
+        if let Some(ref retry_topic) = consumer_config.kafka_retry_topic {
+            topics_to_consume.push(retry_topic.clone());
+        }
```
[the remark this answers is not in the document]
untitaker: Just don't configure it that way. Not my problem!

src/store/adapters/sqlite.rs
```diff
@@ -683,26 +696,65 @@ impl InflightActivationStore for SqliteActivationStore {
...
-        .fetch_optional(&mut *conn)
+        .fetch_optional(&mut *tx)
         .await?;
 
-        let Some(row) = result else {
+        let Some(mut row) = result else {
             return Ok(None);
         };
 
+        let mut activation = TaskActivation::decode(&row.activation as &[u8])?;
```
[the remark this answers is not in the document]
untitaker: > If the activation blob is corrupt or not a valid TaskActivation

what if the sky was pink?

PR #634, opened 2026-05-13
feat(retries):  Add query-level retry for transient db errors

victoria-yining-huang:
ticket https://linear.app/getsentry/issue/STREAM-940/handle-server-shutting-down-errors-and-retry

 - Add retry_query utility function that retries individual database queries on transient infrastructure errors (Io, PoolTimedOut)
  - Errors like "Unable to update status of activation: server conn crashed?" and "Unable to write to sqlite: server shutting down" are now handled gracefully — the query is retried instead
  of immediately surfacing the error
  - Permanent errors (constraint violations, schema errors, PoolClosed, WorkerCrashed) are not retried
  - Retry is applied at the query execution level inside the Postgres adapter (~15 query sites), not at the store method level
  - Two new config options: db_query_max_retries (default None / disabled) and db_query_retry_delay_ms (default 100ms)
  - SQLite adapter is unchanged — pull-mode taskbrokers are unaffected

Methods wrapped with retry_query (hot path — gRPC, push, fetch)

  - store, claim_activations, set_status, set_status_batch, mark_activation_processing, mark_completed, set_processing_deadline, delete_activation
  - db_size, get_by_id, count_by_status, count, count_depths, get_retry_activations

  Methods NOT wrapped (in both adapters)

  - handle_claim_expiration, handle_processing_deadline, handle_processing_attempts, handle_expires_at, handle_delay_until, handle_failed_tasks, remove_completed, remove_killswitched — all
  called by upkeep loop every 1s, which provides natural retry
  - vacuum_db / full_vacuum_db — maintenance, retried by upkeep/maintenance loops
  - pending_activation_max_lag — best-effort metric, returns f64 not Result
  - assign_partitions — not a DB query
  - clear, remove_db — test-only

  Performance note

  The store method passes &rows by reference to push_values and binds fields by reference where possible, avoiding a full Vec<TableRow> clone on every call. SQLite adapter is completely
  untouched — zero overhead for pull-mode taskbrokers.

  Observability

  - store.retry.attempt — counts each retry attempt, tagged by method
  - store.retry.succeeded — query succeeded after retry
  - store.retry.exhausted — all retries exhausted on retryable errors (does not fire if a non-retryable error follows a retryable one)


src/store/retry.rs
```diff
@@ -0,0 +1,601 @@
...
+        Some(sqlx::Error::PoolClosed) => true,
+        Some(sqlx::Error::WorkerCrashed) => true,
+        _ => false,
+    }
+}
+
+/// Calculates the backoff duration for a given attempt using exponential backoff.
+fn backoff_duration(attempt: u32) -> Duration {
+    let ms = BASE_BACKOFF_MS
+        .saturating_mul(1 << attempt)
```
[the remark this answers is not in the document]
victoria-yining-huang: this will only happen if attempt is higher than 63, no one will set that retry number. I also added a comment in config.rs noting this

src/store/retry.rs
```diff
@@ -0,0 +1,601 @@
...
+                            "method" => stringify!($method),
+                        )
+                        .increment(1);
+                    }
+                    return Err(err);
+                }
+            }
+        }
+    }};
+}
```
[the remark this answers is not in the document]
markstory: :+1: to this being more duplication than we should take on. Perhaps there is a solution where the `retry_method` macro can have different signatures for owned/ref parameters, and the clone/ref can be conditionally performed?

src/config.rs
```diff
@@ -180,6 +180,11 @@ pub struct Config {
     /// The amount of time to wait before retrying writes to db when write fails.
     pub db_write_failure_backoff_ms: u64,
 
+    /// The maximum number of times to retry a transient database query error
+    /// before surfacing the error. Uses exponential backoff between retries.
+    /// When None, queries are not retried. Must not exceed 63 retries although something else must be wrong if it goes that high.
```
markstory: Why is 63 relevant? All your ints are u32 which supports values much larger than 63.

victoria-yining-huang: (this is removed now as i removed exponential backoff entirely, just static sleep now) but it was because the variable that holds the duration in ms could have overflow, if the retry number was too high, due to exponential math

victoria-yining-huang: bc transient db errors wouldn't benefit from exponential backoff too much anyway

george-sentry (commented):
I agree that it would be preferable to avoid having two nearly identical macro definitions. Otherwise I like this solution!

src/store/retry.rs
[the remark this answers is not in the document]
victoria-yining-huang: I removed the duplicated function, only cloning at arg level now

victoria-yining-huang:
@george-sentry I removed the duplicated function, only cloning at arg level now

src/store/retry.rs
```diff
@@ -0,0 +1,532 @@
...
+
+#[async_trait]
+impl InflightActivationStore for RetryStore {
+    async fn store(&self, batch: Vec<InflightActivation>) -> Result<u64, Error> {
+        retry_method!(self, store(batch.clone()))
+    }
+
+    fn assign_partitions(&self, partitions: Vec<i32>) -> Result<(), Error> {
+        self.inner.assign_partitions(partitions)
```
george-sentry: Most methods are retried, but not all, such as this one and `remove_db` and `pending_activation_max_lag`. Is there a reason for that?

src/store/retry.rs
```diff
@@ -0,0 +1,532 @@
...
+        let mut attempt = 0u32;
+        loop {
+            match $self.inner.$method( $($arg),* ).await {
+                Ok(val) => {
+                    if attempt > 0 {
+                        info!(
+                            method = stringify!($method),
+                            attempt,
+                            "Query succeeded after retry"
+                        );
```
fpacifici: Please avoid this logging at info level. DEbug would be fine. 
Info is what we log in production, this can be very very high throughput

src/store/retry.rs
```diff
@@ -0,0 +1,532 @@
...
+                    }
+                    return Err(err);
+                }
+            }
+        }
+    }};
+}
+
+#[async_trait]
+impl InflightActivationStore for RetryStore {
```
fpacifici: I don't think this is the right abstraction level where to put the retry (wrapping the entire store with the retry logic). 

The ActivationStore does more than running queries to the DB. So we cannot just decide to retry the entire method if the method raises a retriable DB error. What if you wanted to retry the DB query but not retry the entire method ?
- Let's say your method had side effects beyond the DB query you do not want to retry ?
- What if the method ran two queries and you wanted to retry only one ?

The design argument more in general:
- the store has more responsibilities than just running DB queries. 
- We want to avoid mixing concerns between abstractions levels to keep the design logically organized and the cognitive overhead low.
- When running the query, there are different categories of errors that can happen. We generally want to retry those that are infrastructure related (disconnection, network, etc.) 
- The infrastructure aspects are hidden as lower as possible (like in the connection pool https://github.com/getsentry/taskbroker/blob/main/src/store/adapters/postgres.rs#L106-L114) so that the layers above do not have to be polluted with infrastructure details.
- As we are retrying only infrastructure related issues, the natural place where to put this logic should be as close to the infra as possible so the application only cares of application level issues.

I would recommend adding an abstraction around the connection pool to execute the queries and perform retry when needed rather than retrying the entire method. 
I am not too familiar with sqlx, I know it  does not manage retries for you:
- You can write a wrapper around the connection pool that performs retry
- sqlx produces futures, I think you can build a simple wrapper to retry them https://github.com/getsentry/symbolicator/blob/f8883320dcc5e8eb6db852620c1d143cacfdf9b9/crates/symbolicator-service/src/download/mod.rs#L387-L409

src/store/adapters/sqlite.rs
```diff
@@ -456,73 +422,127 @@ impl InflightActivationStore for SqliteActivationStore {
...
+                        application,
+                        namespace,
+                        taskname,
+                        on_attempts_exceeded,
+                        bucket
+                    )
+                ",
+            );
+            let query = query_builder
+                .push_values(rows.clone(), |mut b, row: TableRow| {
```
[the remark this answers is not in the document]
victoria-yining-huang: This cost is acceptable since retries will be enabled in production.

src/store/adapters/postgres.rs
```diff
@@ -419,122 +432,134 @@ impl InflightActivationStore for PostgresActivationStore {
         bucket: Option<BucketRange>,
```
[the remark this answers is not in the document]
victoria-yining-huang: the temporarily orphaned activation will be handled by upkeep after the expiration has passed, this already happens today when a taskbroker crashes after an activation is claimed. This is OK imo so still worth wrapping it in retry

src/store/adapters/sqlite.rs
```diff
@@ -456,55 +413,111 @@ impl InflightActivationStore for SqliteActivationStore {
...
+            let mut conn = self.acquire_write_conn_metric("store").await?;
+            let result = query.execute(&mut *conn).await?;
+            Ok(result.rows_affected())
+        })
+        .await?;
 
         // Sync the WAL into the main database so we don't lose data on host failure.
+        // Best-effort — errors are swallowed and only recorded as metrics.
         let checkpoint_timer = Instant::now();
+        let mut conn = self.acquire_write_conn_metric("store_checkpoint").await?;
```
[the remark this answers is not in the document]
victoria-yining-huang: this existed before my change, i just copy pasted the same code over, not within the scope of this PR to fix

src/store/adapters/sqlite.rs
```diff
@@ -456,55 +413,111 @@ impl InflightActivationStore for SqliteActivationStore {
...
```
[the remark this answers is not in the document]
victoria-yining-huang: didn't i reply to this already? This code existed beofre my change i just copied it over. No new behaviour change. My understanding is that `ON CONFLICT(id) DO NOTHING` handled this scenario already

src/store/retry.rs
victoria-yining-huang: i followed symbolicator's retry wrapper that you linked and added that at each appropriate call site.

src/store/adapters/postgres.rs
```diff
@@ -339,74 +348,78 @@ impl InflightActivationStore for PostgresActivationStore {
...
+                    }
+                    b.push_bind(row.status);
+                    b.push_bind(row.at_most_once);
+                    b.push_bind(row.application);
+                    b.push_bind(row.namespace);
+                    b.push_bind(row.taskname);
+                    b.push_bind(row.on_attempts_exceeded as i32);
+                    b.push_bind(row.bucket);
+                })
+                .push(" ON CONFLICT(id) DO NOTHING")
```
fpacifici, line 428: It seems to me we are doing something very similar to what we did before.
The `retry_query` call is wrapping mostly the whole method rather than the query execution. 
Is this a requirement ? (like you cannot reuse the QueryBuilder) If not, is it possible to run only the `execute method in the `retry_query` function ?

src/store/adapters/postgres.rs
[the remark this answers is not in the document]
fpacifici: The bot is not wrong in principle. If there was a possibility to have a transient error after commit we could not be running the retry as we are. Letting upkeep take care of scenarios with transient error is not ok, it adds latency for scenarios we are supposed to handle.

Though here nothing happens between commit and return.

src/store/adapters/postgres.rs
```diff
@@ -638,20 +672,23 @@ impl InflightActivationStore for PostgresActivationStore {
...
-        let Some(row) = result else {
-            return Ok(None);
-        };
+            let Some(row) = result else {
+                return Ok(None);
+            };
 
-        Ok(Some(row.into()))
+            Ok(Some(row.into()))
+        })
```
fpacifici: What happens here if we 
- attempt the query, the DB applies the change but a connection error prevents us from getting the result
- retry
Do we get the row back if no change is applied the second time? does the client depend on that ?

src/store/adapters/postgres.rs
```diff
@@ -661,104 +698,132 @@ impl InflightActivationStore for PostgresActivationStore {
...
     // Used in tests
     #[framed]
     async fn clear(&self) -> Result<(), Error> {
-        let mut conn = self.acquire_write_conn_metric("clear").await?;
-        sqlx::query("TRUNCATE TABLE inflight_taskactivations")
-            .execute(&mut *conn)
-            .await?;
-
-        Ok(())
+        retry_query(&self.config.retry_config, "clear", || async {
```
fpacifici: You do not need to retry the operation to wipe the DB

src/store/retry.rs
```diff
@@ -0,0 +1,215 @@
...
+
+/// Returns true if the error is a transient database/connection error
+/// that is likely to succeed on retry. Downcasts the anyhow::Error to
+/// sqlx::Error to match on structured variants rather than parsing strings.
+fn is_retryable_error(err: &Error) -> bool {
+    matches!(
+        err.downcast_ref::<sqlx::Error>(),
+        Some(sqlx::Error::Io(_))
+            | Some(sqlx::Error::PoolTimedOut)
+            | Some(sqlx::Error::PoolClosed)
```
fpacifici: Does this error recover by retrying ? Don't we have to re-establish the connection pool ?

src/store/adapters/sqlite.rs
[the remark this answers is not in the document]
fpacifici: I don't think we can say this. This will happen at every write on sqlte or postgres. So it may not be acceptable. Also we will not be able to turn this off by configuration by removing the retries.
Please reach out to @george-sentry  and run a load test on this.

src/store/adapters/postgres.rs
victoria-yining-huang: correct, `query_builder` and `query` both cannot be reused, that's why they have to be constructed inside retry_query.

src/store/retry.rs
victoria-yining-huang: you're right, only Io and PoolTimedOut are retryable, removed the rest

src/store/adapters/postgres.rs
victoria-yining-huang: removed

src/store/retry.rs
```diff
@@ -0,0 +1,214 @@
...
+                    attempt,
+                    error = %err,
+                    "Retryable database error, retrying"
+                );
+                metrics::counter!("store.retry.attempt", "method" => label).increment(1);
+                sleep(config.retry_delay).await;
+                attempt += 1;
+            }
+            Err(err) => {
+                if attempt > 0 {
```
evanh: Nit: This should also check if the error is retryable. Otherwise it's possible we could get a retryable error, and then a non-retryable error, and this metric would still get emitted.

src/store/adapters/postgres.rs
```diff
@@ -822,26 +884,33 @@ impl InflightActivationStore for PostgresActivationStore {
...
+
+                if let Ok(query_res) = processing_attempts_result {
+                    return Ok(query_res.rows_affected());
+                }
 
-        Err(anyhow!("Could not update tasks past processing_deadline"))
+                Err(anyhow!("Could not update tasks past processing_deadline"))
+            },
+        )
+        .await
```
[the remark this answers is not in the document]
evanh: This is a valid bug.

src/store/adapters/postgres.rs
evanh: Did some more digging, they can be re-used, but they have to be `reset` in between uses. So we need to rebind all the values anyways.

src/store/retry.rs
victoria-yining-huang: this was a good catch, added a test `test_retryable_then_non_retryable_stops_without_exhausting` for it too

src/store/adapters/postgres.rs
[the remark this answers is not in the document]
victoria-yining-huang: i decided to not wrap any methods that are called by upkeep loop for consistency. So `handle_processing_attempts` is not wrapped with retry anymore. If we decide to surface specific `sqlx::Error` that can be done in a subsequent PR and then it becomes doable for retries, but that's a separate decision (retry on operations called by upkeep)

evanh (approved):
Looks good!

Issue #610, opened 2026-04-29
Add structured worker failure fields to the existing task execution log


s-starostin:
This is a narrower follow-up to [#114010](https://github.com/getsentry/sentry/issues/114010), scoped down to worker-local observability only.

Per prior feedback, I’m opening an issue first to get alignment on the path to take before sending any PR.

## Context

My understanding is that exception context is intended to stay with the worker rather than be propagated through taskbroker.

Looking at `clients/python/src/taskbroker_client/worker/workerchild.py`, what I found in the non-retriable failure path is `sentry_sdk.capture_exception(err)` plus the existing task execution status logging. What I could not find there is an accompanying stdout log line with task identifiers and exception details in the same caught-exception path. If I’m reading that incorrectly, please correct me.

## Problem

When debugging starts from task/taskbroker failure symptoms, the worker-local exception context is not always visible in the place the operator is looking.

In a recent incident, the visible symptom was a task ending in `Failure`, while the actual cause only became obvious after manually replaying the failing worker path:

`OperationalError: could not access file "$libdir/btree_gist": No such file or directory`

If that exception was already captured to Sentry, the operational gap for me was still at the `kubectl logs` / worker-log level during incident response.

## Proposed path

Would it be acceptable to extend the existing worker log path with structured failure fields, for example on the existing `taskworker.task_execution` failure log or an adjacent `taskworker.task_failed` log, including:

- `task_id`
- `task_name`
- `namespace`
- `exception_type`
- `exception_message`

Example shape:

`taskworker.task_execution status=failure task_id=... task_name=... namespace=... exception_type=OperationalError exception_message='could not access file "$libdir/btree_gist": No such file or directory'`

## Non-goals

This is intentionally not proposing:

- protobuf changes
- taskbroker payload changes
- broker-side error propagation
- inflight-store persistence changes

## Alignment question

If this is aligned with the intended architecture, I can send a small PR limited to that worker-local logging change. If not, I’m happy to close this and handle it with a runbook entry on our side.

s-starostin:
@markstory, this follows your suggestion to open an issue first and get alignment on the path before sending another PR.

markstory:
> Would it be acceptable to extend the existing worker log path with structured failure fields, for example on the existing taskworker.task_execution failure log or an adjacent taskworker.task_failed log, including:

[...]

PR #635, opened 2026-05-13
feat(worker): add taskworker.task.failed log for failed tasks

s-starostin:
Adds a structured `taskworker.task.failed` log event when task execution fails in
`run_worker`.

For non-retriable failure paths, this replaces direct
`sentry_sdk.capture_exception(err)` calls with:

```python
logger.exception("taskworker.task.failed", extra={...})
```

inside the existing `isolation_scope` blocks where they exist. This gives
operators a structured stdout log with the task identity and original exception
details. In environments where the Sentry SDK `LoggingIntegration` captures
error logs, the same call also produces the corresponding Sentry event.

The log includes:

* `task_id`
* `taskname`
* `namespace`
* `processing_pool`
* `exception_type`
* `exception_message`

## Failure paths covered

* `ProcessingDeadlineExceeded`, gated by `task_func.report_timeout_errors`
* General exception fallback, gated by `task_func.silenced_exceptions` and retry state
* Retry-exhausted (unconditional, per #627 - retry exhaustion is a terminal state worth surfacing even for silenced exception types)

On the retry-exhausted path, the structured log uses the original task exception
(`err`) and fires alongside the existing synthetic `NoRetriesRemainingError`
capture. This keeps retry-exhausted failures visible in worker logs in the same
shape as other task failures, while preserving the existing retry-exhaustion
Sentry grouping/fingerprint.

## Existing behavior preserved

* `task_func.report_timeout_errors` opt-out continues to silence both the
  Sentry-reporting path and the new structured worker log on the timeout path
* `task_func.silenced_exceptions` opt-out continues to silence both surfaces on the general-exception path (non-fexhausted). Retry-exhausted is intentionally unconditional per #627.
* `NoRetriesRemainingError` synthetic Sentry capture and fingerprint are unchanged
* Ambient Sentry scope on the general-exception fallback is unchanged; the call
  is not wrapped in a new `isolation_scope`

## Note on retry-exhausted reporting

When Sentry SDK `LoggingIntegration` is configured to capture error-level logs
(the default), retry-exhausted failures will now produce two Sentry events:

1. the existing synthetic `NoRetriesRemainingError` event for retry-exhaustion
   alerting/grouping
2. an event for the original exception from `taskworker.task.failed`

These represent different exception objects and have distinct grouping behavior.
The synthetic event preserves the existing retry-exhaustion signal, while the
original-exception event/log makes the underlying failure visible to operators.

Refs #610


clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -281,27 +306,30 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
-                            scope.fingerprint = [
-                                "taskworker.no_retries_remaining",
-                                inflight.activation.namespace,
-                                inflight.activation.taskname,
-                            ]
-                            scope.set_transaction_name(inflight.activation.taskname)
-                            sentry_sdk.capture_exception(retry_error)
-                            # In this branch, all exceptions should be either
-                            #  captured or silenced.
+                            if should_capture_error:
```
markstory: Why is this being introduced? This change was recently reverted in #627

clients/python/tests/worker/test_worker.py
```diff
@@ -1122,5 +1150,138 @@ def test_child_process_retry_on_deadline_exceeded(mock_capture: mock.Mock) -> No
...
+@mock.patch("taskbroker_client.worker.workerchild.logger")
+def test_child_process_silenced_retry_exhausted_does_not_log_task_failed(
+    mock_logger: mock.Mock,
+    mock_capture: mock.Mock,
+) -> None:
+    """
+    When err is silenced AND retries are exhausted, neither the synthetic
+    NoRetriesRemainingError capture nor the structured taskworker.task.failed
+    log fires. The silenced_exceptions opt-out covers both surfaces on the
+    retry-exhausted path.
```
markstory: I don't think we need this test, the behavior for silenced errors shouldn't be changed.

s-starostin: Agreed, removed.

clients/python/src/taskbroker_client/worker/workerchild.py
s-starostin: I missed #627. Restored the unconditional capture and dropped the gate.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -292,16 +317,20 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
                             ]
                             scope.set_transaction_name(inflight.activation.taskname)
                             sentry_sdk.capture_exception(retry_error)
-                            # In this branch, all exceptions should be either
-                            #  captured or silenced.
-                            captured_error = True
+                            # Also emit structured stdout log for retry-exhausted failures.
+                            # Uses the original err, not the synthetic retry_error, so the log
+                            # carries the actual exception that exhausted retries.
+                            _log_task_failed(inflight.activation, err, processing_pool_name)
```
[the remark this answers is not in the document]
s-starostin: Bugbot is right that the PR description is stale, but I think the suggested code change is wrong. Retry-exhausted reporting is intentionally unconditional per #627, so I’ll update the PR description instead of gating `_log_task_failed`.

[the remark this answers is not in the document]
markstory: We'll capture two errors here as we have both a `capture_exception()` and `logging.exception()`. Could we skip the logging call here, or only log the `retry_error`, and have the logger be aware of nested errors so that the original type is logged?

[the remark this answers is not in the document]
s-starostin: Good call. I changed this branch to keep `sentry_sdk.capture_exception(retry_error)` as the only error capture, and emit the structured worker log with `logger.warning()` instead of `logger.exception()`.

That keeps the Sentry event centered on `NoRetriesRemainingError`, with the original task exception attached as `__cause__`, while still putting the original exception type/message and retry metadata into the structured stdout log.

markstory (approved):

markstory:
Thank you 🎉

PR #637, opened 2026-05-14
feat(taskbroker): Batch Claimed → Processing Updates

george-sentry:
## Linear
Completes [STREAM-920](https://linear.app/getsentry/issue/STREAM-920/batch-claim-processing-updates)

## Description
Builds on #618, which introduced a generic "flusher" mechanism and configurable batched status updates (that is, status updates triggered by calls to the `SetTaskStatus` RPC endpoint).

Here, we use the same mechanism to batch claimed → processing updates, which I call "push" updates for short (hence names like `push_update_batch_size`, `push_update_tx`, and so on). It can be turned on and off using the `batch_push_updates` boolean flag.

Batching these updates should result in significantly higher throughput (several thousand tasks per second higher, depending on the workload and database).

Concerning data loss, _what if taskbroker crashes while claimed → processing updates are queued_? Those updates are lost, meaning even though those tasks were claimed and sent successfully, they weren't updated to processing. Three things can happen here...

1. If the worker completes a task fast enough, it will be updated to "completed" directly from "claimed" and later removed by upkeep ✅
2. If the worker doesn't complete a task fast enough, it may be reset back to "pending" before the worker completes it. In that case, it will be updated from "pending" directly to "completed" and later removed by upkeep ✅
3. If the worker doesn't complete a task fast enough _and_ the taskbroker has enough time to reclaim and / or resend it again, the task will be completed twice ✅ ✅

In all cases, tasks aren't dropped but may be repeated.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -317,10 +317,10 @@ pub struct Config {
     /// Maximum milliseconds to wait before flushing a batch of dispatch updates.
     pub push_update_interval_ms: u64,
 
-    /// The hostname used to construct `callback_url` for task push requests.
+    /// (DEPRECATED) The hostname used to construct `callback_url` for task push requests.
     pub callback_addr: String,
 
-    /// The port used to construct `callback_url` for task push requests.
+    /// (DEPRECATED) The port used to construct `callback_url` for task push requests.
     pub callback_port: u32,
 
     /// Maps every application to its worker endpoint, both represented as strings.
--- a/src/config.rs
+++ b/src/config.rs
@@ -308,6 +308,15 @@ pub struct Config {
     /// Maximum milliseconds to wait before flushing a batch of status updates.
     pub status_update_interval_ms: u64,
 
+    /// Update claimed → processing (dispatch) updates in batches?
+    pub batch_push_updates: bool,
+
+    /// The size of a batch of dispatch updates.
+    pub push_update_batch_size: usize,
+
+    /// Maximum milliseconds to wait before flushing a batch of dispatch updates.
+    pub push_update_interval_ms: u64,
+
     /// The hostname used to construct `callback_url` for task push requests.
     pub callback_addr: String,
 
@@ -421,6 +430,9 @@ impl Default for Config {
             batch_status_updates: false,
             status_update_batch_size: 1,
             status_update_interval_ms: 100,
+            batch_push_updates: false,
+            push_update_batch_size: 1,
+            push_update_interval_ms: 100,
             callback_addr: "0.0.0.0".into(),
             callback_port: 50051,
             worker_map: [("sentry".into(), "http://127.0.0.1:50052".into())].into(),
--- a/src/config.rs
+++ b/src/config.rs
@@ -222,6 +222,11 @@ pub struct Config {
     /// brokers are under load, or there are small networking delays.
     pub processing_deadline_grace_sec: u64,
 
+    /// The number of additional seconds that claim expirations
+    /// are extended by. This helps reduce claim expirations when
+    /// brokers are under load, or there are small networking delays.
+    pub claim_expiration_grace_sec: u64,
+
     /// The frequency at which upkeep tasks
     /// (discarding, retrying activations, etc.) are executed.
     pub upkeep_task_interval_ms: u64,
@@ -406,6 +411,7 @@ impl Default for Config {
             max_processing_count: 2048,
             max_processing_attempts: 5,
             processing_deadline_grace_sec: 3,
+            claim_expiration_grace_sec: 3,
             upkeep_task_interval_ms: 1000,
             upkeep_unhealthy_interval_ms: 5000,
             health_check_killswitched: false,
--- a/src/config.rs
+++ b/src/config.rs
@@ -31,6 +31,12 @@ pub enum DeliveryMode {
     Push,
 }
 
+impl DeliveryMode {
+    pub fn is_push(self) -> bool {
+        self == DeliveryMode::Push
+    }
+}
+
 #[derive(PartialEq, Debug, Deserialize, Serialize)]
 pub struct Config {
     /// The sentry DSN to use for error reporting.
@@ -304,7 +310,7 @@ pub struct Config {
     /// Maximum time in milliseconds for a single push RPC to the worker service. This should be greater than the worker's internal timeout.
     pub push_timeout_ms: u64,
 
-    /// Update statuses from the gRPC server in batches?
+    /// Update statuses from the gRPC server in batches? Only applies in PUSH mode.
     pub batch_status_updates: bool,
 
     /// The size of a batch of status updates.
@@ -313,7 +319,7 @@ pub struct Config {
     /// Maximum milliseconds to wait before flushing a batch of status updates.
     pub status_update_interval_ms: u64,
 
-    /// Update claimed → processing (dispatch) updates in batches?
+    /// Update claimed → processing (dispatch) updates in batches? Only applies in PUSH mode.
     pub batch_push_updates: bool,
 
     /// The size of a batch of dispatch updates.
--- a/src/fetch/mod.rs
+++ b/src/fetch/mod.rs
@@ -122,11 +122,10 @@ impl<T: TaskPusher + Send + Sync + 'static> FetchPool<T> {
                         }
 
                         _ = async {
-                            let start = Instant::now();
-
                             debug!("Fetching next batch of pending activations...");
                             metrics::counter!("fetch.loop.count").increment(1);
 
+                            let start = Instant::now();
                             let mut backoff = false;
 
                             let result = store.claim_activations_for_push(limit, bucket).await;
--- a/src/fetch/mod.rs
+++ b/src/fetch/mod.rs
@@ -48,7 +48,8 @@ pub fn bucket_range_for_fetch_thread(thread_index: usize, fetch_threads: usize) 
     (low, high)
 }
 
-/// Thin interface for the push pool. It mostly serves to enable proper unit testing, but it also decouples fetch logic from push logic even further.
+/// Thin interface for the push pool. It mostly serves to enable proper unit testing,
+/// but it also decouples fetch logic from push logic even further.
 #[async_trait]
 pub trait TaskPusher {
     /// Submit a single task to the push pool.
@@ -164,7 +165,13 @@ impl<T: TaskPusher + Send + Sync + 'static> FetchPool<T> {
                                             )
                                             .record(latency as f64);
                                         } else {
-                                            debug!(task_id = %id, namespace = activation.namespace, taskname = activation.taskname, "Activation already processed, skipping received → claimed latency recording");
+                                            debug!(
+                                                task_id = %id,
+                                                namespace = activation.namespace,
+                                                taskname = activation.taskname,
+                                                "Activation already processed, skipping \
+                                                 received → claimed latency recording"
+                                            );
                                         }
 
                                         match pusher.submit_task(activation, start).await {
--- a/src/fetch/tests.rs
+++ b/src/fetch/tests.rs
@@ -98,8 +98,12 @@ impl InflightActivationStore for MockStore {
         })
     }
 
-    async fn mark_activation_processing(&self, _id: &str) -> Result<(), Error> {
+    async fn mark_processing(&self, _id: &str) -> Result<(), Error> {
         Ok(())
+    }
+
+    async fn mark_processing_batch(&self, _ids: &[String]) -> Result<u64, Error> {
+        unimplemented!()
     }
 
     async fn pending_activation_max_lag(&self, _now: &DateTime<Utc>) -> f64 {
--- a/src/flusher.rs
+++ b/src/flusher.rs
@@ -29,7 +29,8 @@ where
 
     loop {
         tokio::select! {
-            msg = rx.recv() => {
+            // When the buffer is NOT full, try to receive another message
+            msg = rx.recv(), if buffer.len() < batch_size => {
                 match msg {
                     Some(v) => {
                         buffer.push(v);
@@ -51,6 +52,19 @@ where
                 }
             }
 
+            // If the buffer IS full, the branch above will never execute, and we will never
+            // discover that the channel is now closed, which is why this branch is necessary
+            _ = std::future::ready(()), if rx.is_closed() => {
+                while let Ok(update) = rx.try_recv() {
+                    // Buffer may grow beyond configured limit, which is OK because we are shutting down
+                    buffer.push(update);
+                }
+
+                flush(&mut buffer).await;
+                break;
+            }
+
+            // Otherwise, try flushing whatever is in the buffer every `interval_ms` milliseconds
             _ = interval.tick() => {
                 if !buffer.is_empty() {
                     flush(&mut buffer).await;
--- a/src/flusher.rs
+++ b/src/flusher.rs
@@ -3,6 +3,7 @@ use std::pin::Pin;
 use std::time::Duration;
 
 use anyhow::Result;
+use elegant_departure::get_shutdown_guard;
 use tokio::sync::mpsc::Receiver;
 use tracing::debug;
 
@@ -28,6 +29,8 @@ where
 
     let mut buffer: Vec<T> = Vec::with_capacity(batch_size);
 
+    let guard = get_shutdown_guard().shutdown_on_drop();
+
     loop {
         tokio::select! {
             biased;
@@ -51,38 +54,37 @@ where
                     }
 
                     None => {
-                        // Channel closed (shutdown), flush remaining and exit
-                        debug!("Channel closed due to shutdown, flushing remaining before exit...");
-                        flush(&mut buffer).await;
+                        // Channel closed
+                        debug!("Channel closed!");
                         break;
                     }
                 }
-            }
-
-            // If the buffer IS full, the branch above will never execute, and we will never
-            // discover that the channel is now closed, which is why this branch is necessary
-            _ = std::future::ready(()), if rx.is_closed() => {
-                debug!("Channel is closed and buffer is full, draining channel before exiting...");
-
-                while let Ok(update) = rx.try_recv() {
-                    // Buffer may grow beyond configured limit, which is OK because we are shutting down
-                    buffer.push(update);
-                }
-
-                flush(&mut buffer).await;
-                break;
             }
 
             // Otherwise, try flushing whatever is in the buffer every `interval_ms` milliseconds
             _ = interval.tick() => {
                 debug!("Performing periodic flush...");
 
-                if !buffer.is_empty() {
-                    flush(&mut buffer).await;
+                if rx.is_closed() {
+                    debug!("Channel closed on tick!");
+                    break;
                 }
+
+                flush(&mut buffer).await;
+            }
+
+            _ = guard.wait() => {
+                debug!("Shutdown guard triggered!");
+                break;
             }
         }
     }
 
+    // Drain and flush before exit
+    while let Ok(update) = rx.try_recv() {
+        buffer.push(update);
+    }
+
+    flush(&mut buffer).await;
     Ok(())
 }
--- a/src/flusher.rs
+++ b/src/flusher.rs
@@ -4,6 +4,7 @@ use std::time::Duration;
 
 use anyhow::Result;
 use tokio::sync::mpsc::Receiver;
+use tracing::debug;
 
 /// Run flusher that receives values of type T from a channel and flushes
 /// them using the provided async `flush` function either when the batch is
@@ -31,6 +32,8 @@ where
         tokio::select! {
             // When the buffer is NOT full, try to receive another message
             msg = rx.recv(), if buffer.len() < batch_size => {
+                debug!("Buffer is NOT full, receiving a message...");
+
                 match msg {
                     Some(v) => {
                         buffer.push(v);
@@ -40,12 +43,14 @@ where
                         }
 
                         if buffer.len() >= batch_size {
+                            debug!("Flushing full buffer...");
                             flush(&mut buffer).await;
                         }
                     }
 
                     None => {
                         // Channel closed (shutdown), flush remaining and exit
+                        debug!("Channel closed due to shutdown, flushing remaining before exit...");
                         flush(&mut buffer).await;
                         break;
                     }
@@ -55,6 +60,8 @@ where
             // If the buffer IS full, the branch above will never execute, and we will never
             // discover that the channel is now closed, which is why this branch is necessary
             _ = std::future::ready(()), if rx.is_closed() => {
+                debug!("Channel is closed and buffer is full, draining channel before exiting...");
+
                 while let Ok(update) = rx.try_recv() {
                     // Buffer may grow beyond configured limit, which is OK because we are shutting down
                     buffer.push(update);
@@ -66,6 +73,8 @@ where
 
             // Otherwise, try flushing whatever is in the buffer every `interval_ms` milliseconds
             _ = interval.tick() => {
+                debug!("Performing periodic flush...");
+
                 if !buffer.is_empty() {
                     flush(&mut buffer).await;
                 }
--- a/src/flusher.rs
+++ b/src/flusher.rs
@@ -30,6 +30,8 @@ where
 
     loop {
         tokio::select! {
+            biased;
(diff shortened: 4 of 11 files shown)


src/fetch/mod.rs
```diff
@@ -122,11 +122,10 @@ impl<T: TaskPusher + Send + Sync + 'static> FetchPool<T> {
                         }
 
                         _ = async {
-                            let start = Instant::now();
-
                             debug!("Fetching next batch of pending activations...");
                             metrics::counter!("fetch.loop.count").increment(1);
 
+                            let start = Instant::now();
```
george-sentry, line 129: Starting the timer here results in more accurate metrics. It was already like this before but I unintentionally moved it in #618.

src/fetch/tests.rs
```diff
@@ -98,10 +98,14 @@ impl InflightActivationStore for MockStore {
         })
     }
 
-    async fn mark_activation_processing(&self, _id: &str) -> Result<(), Error> {
+    async fn mark_processing(&self, _id: &str) -> Result<(), Error> {
```
george-sentry, line 101: Minor naming change to match `mark_completed` and shorten things.

src/push/mod.rs
```diff
@@ -244,11 +257,31 @@ impl PushPool {
                                         }
 
                                         let start = Instant::now();
-                                        let result = store.mark_activation_processing(&id).await;
-                                        metrics::histogram!("push.mark_activation_processing.duration").record(start.elapsed());
+
+                                        // Are we batching claimed → processing updates?
+                                        if let Some(ref tx) = update_tx {
+                                            let result = tx.send(id.clone()).await;
+                                            metrics::histogram!("push.mark_processing.duration").record(start.elapsed());
```
george-sentry: Even though we aren't calling `mark_processing` here, I think it would be nice to emit the same metric as when we _do_ call that method. It will allow for easier comparison between batched and single updates.

george-sentry: In other words, that naming is intentional.

src/config.rs
```diff
@@ -308,6 +308,15 @@ pub struct Config {
...
     pub status_update_interval_ms: u64,
 
+    /// Update claimed → processing (dispatch) updates in batches?
+    pub batch_push_updates: bool,
+
+    /// The size of a batch of dispatch updates.
+    pub push_update_batch_size: usize,
+
+    /// Maximum milliseconds to wait before flushing a batch of dispatch updates.
+    pub push_update_interval_ms: u64,
```
george-sentry, line 329: As soon as all this is done, we should probably reorganize the configuration. It has become enormous 😅

src/flusher.rs
```diff
@@ -39,19 +43,38 @@ where
...
+            _ = std::future::ready(()), if rx.is_closed() => {
+                debug!("Channel is closed and buffer is full, draining channel before exiting...");
+
+                while let Ok(update) = rx.try_recv() {
+                    // Buffer may grow beyond configured limit, which is OK because we are shutting down
+                    buffer.push(update);
+                }
+
+                flush(&mut buffer).await;
+                break;
```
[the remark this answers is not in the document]
george-sentry: We wouldn't want to retry failed updates during shutdown, as that would delay shutdown.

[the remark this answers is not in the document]
george-sentry: Note that Vic is currently working on a way to retry transient database errors anyways, so this isn't a problem.

src/push/mod.rs
```diff
@@ -422,5 +459,55 @@ async fn push_task(
     result
 }
 
+pub async fn flush_updates(store: Arc<dyn InflightActivationStore>, buffer: &mut Vec<String>) {
+    if buffer.is_empty() {
+        return;
+    }
+
+    let start = Instant::now();
+    let ids = std::mem::take(buffer);
```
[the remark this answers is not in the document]
george-sentry: The formatter prefers this approach.

src/main.rs
```diff
@@ -265,8 +265,30 @@ async fn main() -> Result<(), Error> {
         }
     });
 
+    // Push update flush task
+    let (push_update_tx, push_update_task) = if config.batch_push_updates {
+        let (tx, rx) = tokio::sync::mpsc::channel(config.push_update_batch_size.max(1));
```
enochtangg: What is the reasoning behind setting the channel size to the batch flush size? Did we do any testing around how changing the amount slack between these two configurations affect the system?

enochtangg: A metric that would be interesting is the channel's depth which might give some intuition behind if producers want to send faster than the flusher drains

enochtangg (commented):
Looks good, great work! I had one comment below.

src/push/mod.rs
```diff
@@ -304,18 +337,37 @@ impl PushPool {
                                 debug!(task_id = %id, "Activation sent to worker");
 
                                 let start = Instant::now();
-                                let result = store.mark_activation_processing(&id).await;
-                                metrics::histogram!("push.mark_activation_processing.duration")
-                                    .record(start.elapsed());
 
-                                if let Err(e) = result {
-                                    metrics::counter!("push.mark_activation_processing", "result" => "error").increment(1);
+                                if let Some(ref tx) = update_tx {
```
markstory: There are two similar passages here, could they be refactored into a shared method?

src/push/mod.rs
```diff
@@ -422,5 +474,55 @@ async fn push_task(
...
+    }
+
+    let start = Instant::now();
+    let ids: Vec<_> = std::mem::take(buffer);
+
+    let requested = ids.len() as u64;
+    metrics::histogram!("push.flush_updates.requested").record(requested as f64);
+
+    let result = store.mark_processing_batch(&ids).await;
+    metrics::histogram!("push.mark_processing_batch.duration").record(start.elapsed());
```
markstory, line 433: :+1: This will make it easy to see the impact on latency that batching updates has.

src/config.rs
markstory: We could break it up into a few different nested structures.

src/flusher.rs
```diff
@@ -39,19 +43,38 @@ where
...
+                        debug!("Channel closed due to shutdown, flushing remaining before exit...");
                         flush(&mut buffer).await;
                         break;
                     }
                 }
             }
 
+            // If the buffer IS full, the branch above will never execute, and we will never
+            // discover that the channel is now closed, which is why this branch is necessary
+            _ = std::future::ready(()), if rx.is_closed() => {
```
markstory: Should this select be `biased` to ensure the buffer length and rx are always checked first?

src/main.rs
george-sentry: Good point. Added gauges for `grpc_server.update_queue.depth` and `push.update_queue.depth`.

george-sentry: And no, I haven't extensively tested different channel / batch flush sizes. Good idea for the future.

src/push/mod.rs
george-sentry: Done! In another PR, I would like to actually create a new `PushThread` abstraction because now there are two standalone functions (`push_task` and `send_task`) called from inside this loop that take like half a dozen arguments. But that's for another time.

src/flusher.rs
george-sentry: Yes, made this change.

src/flusher.rs
```diff
@@ -39,25 +48,43 @@ where
...
-                if !buffer.is_empty() {
-                    flush(&mut buffer).await;
+                debug!("Performing periodic flush...");
+
+                if rx.is_closed() {
+                    debug!("Channel closed on tick!");
+                    break;
                 }
+
+                flush(&mut buffer).await;
```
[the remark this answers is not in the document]
george-sentry: I think it's better for `flush` to decide what to do when the buffer is empty.

src/push/mod.rs
```diff
@@ -409,7 +376,7 @@ async fn push_task(
 
     let request = PushTaskRequest {
         task: Some(task),
-        callback_url,
+        callback_url: "".into(),
```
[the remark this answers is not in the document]
george-sentry: We don't use this field anymore, so I'm going to remove it from the schema soon after this PR is completed.

src/push/mod.rs
fpacifici, line 178: Do we have a max line length in our formatter ?

src/push/mod.rs
```diff
@@ -202,135 +210,22 @@ impl PushPool {
 
                                 metrics::histogram!("push.queue.latency").record(time.elapsed());
 
-                                let id = activation.id.clone();
```
fpacifici, line 205: Thanks for getting rid of this copy paste

src/push/mod.rs
```diff
@@ -202,135 +210,22 @@ impl PushPool {
...
-                                        metrics::counter!("push.push_task", "result" => "error").increment(1);
-
-                                        error!(
-                                            task_id = %id,
-                                            error = ?e,
-                                            "Failed to send activation to worker"
-                                        )
-                                    }
-                                };
+                                push_task(store.clone(), update_tx.as_ref(), activation, &mut workers, timeout, grpc_shared_secret.as_slice()).await;
```
fpacifici, line 213: You are passing a number of config attributes to `push_task` like timeout and grpc_shared_secret. Have you considered passing the config struct instead ? Or are you trying to avoid cloning ?

src/push/mod.rs
```diff
@@ -386,12 +281,106 @@ impl PushPool {
...
+                    "taskname" => activation.taskname,
+                )
+                .record(latency as f64);
+            } else {
+                debug!(task_id = %id, namespace = activation.namespace, taskname = activation.taskname, "Activation already processed, skipping received → push latency recording");
+            }
+
+            let start = Instant::now();
+
+            if let Some(tx) = update_tx {
```
fpacifici, line 340: When can this be None ?

src/push/mod.rs
```diff
@@ -386,12 +281,106 @@ impl PushPool {
...
+                metrics::histogram!("push.mark_processing.duration").record(start.elapsed());
+
+                if let Err(e) = result {
+                    metrics::counter!("push.mark_processing", "result" => "error").increment(1);
+
+                    error!(
+                        task_id = %id,
+                        error = ?e,
+                        "Failed to enqueue push update"
+                    );
```
fpacifici, line 354: I'd assume this can happen if the queue is full.
Can this happen under normal operations (let's say we are burning a backlog and the DB is overloaded) ?
How would we be exerting backpressure to prevent us from sending further tasks to the worker if the queue to change their state is full ? we should at least stopping the calls to push_task

src/push/mod.rs
```diff
@@ -386,12 +281,106 @@ impl PushPool {
...
+
+                metrics::histogram!(
+                    "push.received_to_push.latency",
+                    "namespace" => activation.namespace,
+                    "taskname" => activation.taskname,
+                )
+                .record(latency as f64);
+            } else {
+                debug!(task_id = %id, namespace = activation.namespace, taskname = activation.taskname, "Activation already processed, skipping received → push latency recording");
+            }
```
fpacifici, line 336: We were not running this code when draining the receiver. Now this will be added in that scenario. Is it expected ?

src/push/mod.rs
```diff
@@ -386,12 +281,106 @@ impl PushPool {
...
+
+        // Once processing deadline expires, status will be set back to pending
+        Err(e) => {
+            metrics::counter!("push.push_task", "result" => "error").increment(1);
+
+            error!(
+                task_id = %id,
+                error = ?e,
+                "Failed to send activation to worker"
+            )
```
fpacifici, line 380: If we fail to send to the worker we should switch the task back to pending otherwise we would be waiting a long time for tasks that could have been just pushed back for an HTTP 429.

src/main.rs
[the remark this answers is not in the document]
george-sentry: This is actually a good thing.

src/flusher.rs
```diff
@@ -27,9 +29,16 @@ where
 
     let mut buffer: Vec<T> = Vec::with_capacity(batch_size);
 
+    let guard = get_shutdown_guard().shutdown_on_drop();
```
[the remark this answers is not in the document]
george-sentry: Not true. The process will not terminate until all tasks holding the shutdown guard finish, including the gRPC server.

src/push/mod.rs
```diff
@@ -422,5 +412,55 @@ async fn push_task(
...
+        return;
+    }
+
+    let start = Instant::now();
+    let ids: Vec<_> = std::mem::take(buffer);
+
+    let requested = ids.len() as u64;
+    metrics::histogram!("push.flush_updates.requested").record(requested as f64);
+
+    let result = store.mark_processing_batch(&ids).await;
```
fpacifici, line 432: I'd assume we expect Vic's work to provide retries every time we talk to the DB. is this correct ?

src/push/mod.rs
```diff
@@ -422,5 +412,55 @@ async fn push_task(
...
+            error!(
+                requested,
+                error = ?e,
+                "Failed to flush update batch from push pool"
+            );
+
+            // Push failed updates back into the buffer so they can be retried on next flush
+            for id in ids {
+                buffer.push(id);
+            }
```
fpacifici, line 466: Do we cover any scenario other than DB failures here ? If yes what are the scenarios we would cover here that are not covered by Vic's retries ?

src/push/mod.rs
```diff
@@ -422,5 +412,55 @@ async fn push_task(
...
+    metrics::histogram!("push.mark_processing_batch.duration").record(start.elapsed());
+
+    match result {
+        Ok(affected) => {
+            metrics::histogram!("push.flush_updates.affected").record(affected as f64);
+
+            metrics::counter!("push.flush_updates.updated").increment(affected);
+            metrics::counter!("push.flush_updates", "result" => "ok").increment(1);
+
+            if affected < requested {
```
fpacifici, line 442: What would this mean for the application ? Is this a possible scenario ?

src/push/tests.rs
```diff
@@ -451,6 +507,130 @@ async fn push_pool_start_does_not_mark_activation_processing_on_push_failure() {
...
+    let ch_id = update_rx
+        .recv()
+        .await
+        .expect("Task ID should be sent on update channel");
+    assert_eq!(ch_id, id);
+}
+
+/// Function `flush_updates` drains the buffer into `mark_processing_batch` and clears the buffer.
+#[tokio::test]
+async fn flush_updates_applies_batch_and_clears_buffer() {
```
fpacifici, line 564: Do we ever test the scenario where send_task fails to send the request to the GRPC server receiving 429 ?

fpacifici (commented):
Some questions and comments.

Overall some food about the design that is emerging now as we add batching. I noticed it is becoming hard to follow the state machine the task goes through, specifically it is hard to wrap someone's mind about the consistency impact of how we manage specific failure modes. Every time I see an `Err(...)` I need to do quite some search to understand what the impact of that would be. 

I think we should make the state machine explicit in the design rather than making it an emergent property. Some high level ideas on how to do this:
- Have a TaskActivationStateMachine that knows how to apply changes and how to recover from specific events:
- - operations can be: claim, send, fail, revert, fix
- - here is where the business logic sits and it is the place where errors are classified.
- Have three categories of errors: retriable, fatal, recoverable (recoverable, are sent back to the state machine to fix the current state)
- Have an abstraction to batch updates, similar to what you have, though the flush method instantiate the state machine instead of interacting with the store directly. So the code that manages the activation does not need to care whether we update in batches 
- The tasks coordinate the work. 

What do you think about a refactoring like this ?

src/push/mod.rs
george-sentry: This means upkeep moved some of the tasks in that batch from claimed back to pending before we had a chance to mark them as processing. This means claims expire too quickly. This can happen if the claim expiration isn't calculated correctly.

src/push/mod.rs
george-sentry: No, just database failures here. Vic's retries should cover that pretty well.

src/push/mod.rs
george-sentry: Correct.

src/push/mod.rs
george-sentry: In this PR, or another one later?

src/push/mod.rs
george-sentry: Yeah that's okay. I don't think recording this metric will produce a noticeable shutdown delay.

src/push/mod.rs
fpacifici: later is fine. We jsut need to have it in writing to avoid forgetting about it

src/push/mod.rs
fpacifici: Then shouldn't this be covered by Vic retries ?

src/push/mod.rs
george-sentry: When `config.batch_push_updates` is false.

src/push/mod.rs
george-sentry: When the queue is full, it will wait for `tx.send` to finish. I thought about adding a timeout there, but it would create really complicated scenarios like, what do you do when you fail to enqueue an update in X seconds?

That means the only scenario in which this branch will trigger is when the channel is closed.

The channel used to buffer status updates is an MPSC (multiple producer, single consumer) channel. It closes when either (a) all the senders are dropped, or (b) the receiver is dropped.

The **flusher** holds the **receiver**, but the flusher doesn't drop the receiver until the channel is closed. Therefore, the channel will not close until all the **senders** are dropped. In this case, the **push pool** holds the only **sender**. Therefore, this branch should never execute.

src/push/mod.rs
george-sentry: I'll create a Linear ticket for this.

src/push/mod.rs
george-sentry: I'm assuming Vic is implementing a bounded number of retries. If those retries fail, we still need to try these updates again.

src/push/tests.rs
george-sentry: Right now all failures are treated the same, so technically, yes. That would be covered by...

* `push_pool_start_does_not_mark_processing_on_push_failure`
* `push_task_propagates_client_error`

Both use `MockWorkerClient` / `NotifyingWorkerClient` with `should_fail` set to `true`, which returns an error when calling `send` on the worker client.

src/push/mod.rs
george-sentry: Ticket created (STREAM-1006).

src/push/mod.rs
george-sentry: This file is starting to become messy. I want to introduce a `PushThread` abstraction that will make it unnecessary to pass these arguments over and over again.

I could clone `config` in the meantime. Since it's wrapped inside of an `Arc`, the clone would be really cheap so that's not a problem.

Should I do that now, or wait until I create the abstraction? I already have a draft, but I don't want to create a PR for it until this thing is merged.

src/push/mod.rs
george-sentry: Just the default, which is 100. As mentioned on Slack, this probably wasn't shortened because it's indented too far. `cargofmt` can't break this line apart in a way that doesn't go above the line limit, so it just gives up. In these cases, it will be necessary to manually reformat.

src/main.rs
```diff
@@ -265,19 +266,42 @@ async fn main() -> Result<(), Error> {
         }
     });
 
+    // Push update flush task
+    let (push_update_tx, push_update_task) =
+        if config.batch_push_updates && config.delivery_mode.is_push() {
+            let (tx, rx) = tokio::sync::mpsc::channel(config.push_update_batch_size.max(1));
+
```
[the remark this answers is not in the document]
george-sentry: I don't this this is a problem. We can't really shut down gracefully when the database is hanging no matter what. Kubernetes will kill pods after 10 seconds anyways so I'm not concerned.

src/push/mod.rs
george-sentry: It can also happen if claimed activations are executed and marked "completed" before we have a chance to mark them as "processing."

src/flusher.rs
```diff
@@ -39,25 +47,42 @@ where
...
     }
 
+    // Drain and flush before exit
+    while let Ok(update) = rx.try_recv() {
+        buffer.push(update);
+    }
+
+    // Delay shutdown until we have flushed everything in the buffer
+    flush(&mut buffer).await;
+    drop(guard);
```
george-sentry, line 85: The logic is a bit tricky here, but I've confirmed that it works.

The flusher exits the loop when the channel closes. That happens when all senders are dropped. In the case of the status flusher, that's when the gRPC server exits. In the case of the push flusher, that's when the push pool exits.

That's why I don't wait on the shutdown guard here. If I wait on the shutdown guard, I may exit the flusher faster than the gRPC server and push pool. The result will be errors in the gRPC server and push pool since the channel for batching updates is now closed.

By waiting for the channel to close, I avoid that. I also need to make sure I flush all the tasks before exiting. I do this by creating a shutdown guard and dropping it only when the flusher is done. This is necessary because I don't otherwise use the shutdown guard, and shutdown completes only when all guards have been dropped.

I tested this by adding a 30 second sleep after the flush and running it in sandbox. The result was that taskbroker did not shut down completely until the 30 second sleep was complete, as desired.

src/main.rs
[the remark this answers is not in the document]
fpacifici: > Kubernetes will kill pods after 10 seconds anyways so I'm not concerned.

It is more like 5 minutes.

src/push/mod.rs
fpacifici: Maybe cloning the entire config is not a good idea. This happens at every task. Feel free to keep it as it is if the alternative is cloning the whole config.

src/push/mod.rs
fpacifici: ok

src/push/mod.rs
fpacifici: I feel we are becoming a little inconsistent in our error management. 
I think there are three levels:
- basic retriable connection issues -> retried immediately for a fixed number of times -> this is done lower in the stack
- basic connection issues when sending the task to the worker -> like DB connection issues.
- If those are not enough we can assume some larger problems. 
- - Downstream load -> we want backpressure potentially till we can unblock. No restart
- - Same for the overload in pushing the task to the worker
[here there is a spot in between we should discuss]
- Errors from the downstream (DB or worker) that are not overload, that keep happening -> crash and restart

So I have two questions:
- are we consistently applying the policy above? (retry retriable error, backpressure on overload, always reach a point where we restart the broker on repeated errors) ?
- Can we distinguish the scenario between overload and error here ? Piling up work for overlaod is fine, in case of errors, instead we want to be able to crash at a point so that we can recover from bad states.

src/push/mod.rs
fpacifici: > I thought about adding a timeout there, but it would create really complicated scenarios like, what do you do when you fail to enqueue an update in X seconds?

If we do not have a timeout, the broker can hang indefinitely right? If yes we cannot run in production that way. 
The complexity comesfrom the fact that we do not have a clear error management policy I think. See this comment: https://github.com/getsentry/taskbroker/pull/637/changes/BASE..1c31cef0061a3e87b8a1a0e4d6de270088676f94#r3269349218

I would not mind taking a step back for a second and clean this up before moving ahead. If it introduces delay, so be it.

fpacifici: > The channel used to buffer status updates is an MPSC (multiple producer, single consumer) channel. It closes when either (a) all the senders are dropped, or (b) the receiver is dropped.
> 
> The flusher holds the receiver, but the flusher doesn't drop the receiver until the channel is closed. Therefore, the channel will not close until all the senders are dropped. In this case, the push pool holds the only sender. Therefore, this branch should never execute.

Then there should be an assertion or we should let the error bubble up.
Let's pretend, for the sake of the argument, this happens, unless there is a way to recover cleanly the system is in an inconsistent state, thus let it crash. If there was a clean way to recover while leaving the system in a consistent state, we could defensively apply that procedure, but it does not seem to be possible.

src/flusher.rs
fpacifici: This should be in a comment in code. You are explainign why you are not waiting for the guard. Imagine reading this code in a year without context.

george-sentry:
Closing in favor of #652, #653, and others coming soon.

PR #641, opened 2026-05-19
feat(taskbroker): Revamp Configuration

george-sentry:
## Linear
Completes [STREAM-843](https://linear.app/getsentry/issue/STREAM-843/better-configuration-organization)

## Description
The `Config` struct has become enormous due to the various new options introduced by the push and raw delivery modes. It will continue to grow as we work on taskbroker. By splitting the configuration by concern, we make it significantly easier to read and update the configuration, increasing development velocity in the long run.

My goal was to make every component of the application require only the configuration for that specific component, which I achieved for the most part. For example, the `PushPool` takes only `config.push` and the `FetchPool` takes only `config.pull`.

And instead of mixing SQLite and Postgres configuration in one big bag, I separate them into `config.store.pg` and `config.store.sqlite` while `config.store` contains configuration for all adapters. This pattern actually **mirrors what we already have** with `PostgresActivationStoreConfig` and `SqliteActivationStoreConfig`, but there's no need to duplicate anything.

Now, if someone adds config options, it will be obvious what component of taskbroker they influence.

Options are still provided using a combination of YAML and environment variables. To support nested configuration structs, I made double underscores (`__`) represent a period (`.`). For example, to set the PG host, you would set `TASKBROKER_STORE__PG__HOST` or provide the following YAML.

```yaml
store:
  pg:
    host: localhost
```


src/config/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+
+    /// Port for the gRPC server.
+    pub grpc_port: u32,
+
+    /// A list of shared secrets that clients use to authenticate.
+    /// We support a list of secrets to allow for key rotation.
+    pub secrets: Vec<String>,
+
+    /// How often maintenance tasks (reclaiming free pages) are executed.
+    pub maintenance_task_interval_ms: u64,
```
evanh: This should be in the store.rs

src/config/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+
+    /// Enable to have the application perform `VACUUM` on the database
+    /// when it starts up, but before the gRPC server, consumer, and upkeep begin.
+    pub full_vacuum_on_start: bool,
```
evanh: This should be in the store.rs

src/config/mod.rs
```diff
@@ -0,0 +1,829 @@
...
+    /// Build a `Config` instance using (1) defaults, (2) an optional YAML file, and
+    /// (3) environment variables prefixed prefixed with `TASKBROKER_` in that order.
+    pub fn from_args(args: &Args) -> Result<Self, Box<figment::Error>> {
+        let mut builder = Figment::from(Config::default());
+
+        if let Some(path) = &args.config {
+            builder = builder.merge(Yaml::file(path));
+        }
+
+        builder = builder.merge(Env::prefixed("TASKBROKER_").split("__"));
```
evanh, line 143: Does this still work the way we expect? E.g. if we have an env variable `TASKBROKER_MAX_MESSAGE_SIZE` will this still correctly override the `config.kafka.max_message_size`?

PR #642, opened 2026-05-20
feat(client): workerchild awaits TaskProducer futures

bmckerry:
This PR updates the worker child process to:
- pull any pending producer futures created via `TaskProducer`
- store them in memory and check on their status each iteration of the `run_worker` loop
- only send a `ProcessingResult` for a task that has pending futures once all futures in the task are complete
- drain all pending futures on child process shutdown


clients/python/src/examples/README.md
```diff
@@ -24,7 +24,7 @@ With all of those pre-requisites complete, you can run the example application:
 
 ```bash
 # Generate 5 tasks
-python src/examples/cli.py generate --count 5
+python src/examples/cli.py spawn --count 5
```
bmckerry, line 27: Unrelated docs fix, `spawn` is the correct command

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -167,6 +142,18 @@ def child_process(
...
+
+    def handle_sigterm(*args: Any) -> None:
+        logger.info(
+            "taskworker.worker.sigterm_received",
+            extra={"processing_pool": processing_pool_name},
+        )
+        local_shutdown.set()
+
+    signal.signal(signal.SIGTERM, handle_sigterm)
+    signal.signal(signal.SIGINT, signal.SIG_IGN)
```
bmckerry, line 211: SIGINT is already handled in the worker parent process, which uses the standard shutdown path for child processes, ensuring they drain pending futures. This means we can ignore SIGINT in the worker child processes.

clients/python/src/taskbroker_client/worker/workerchild.py
bmckerry, line 323: Broken out into `_task_execution_complete()` below

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -167,6 +142,18 @@ def child_process(
     app = import_app(app_module)
     app.load_modules()
     metrics = app.metrics
+    # Signals when the parent worker pool terminates the child
+    local_shutdown = threading.Event()
```
untitaker, line 201: what is the difference to `shutdown_event`?

bmckerry: If I'm understanding this right, `shutdown_event` is shared across all worker child processes in the workerpool: https://github.com/getsentry/taskbroker/blob/main/clients/python/src/taskbroker_client/worker/worker.py#L633

I didn't want to trigger shutdown on all child processes at once by setting `shutdown_event` in a single child process, so each worker child uses `local_shutdown` when getting SIGTERM.

untitaker: ah `shutdown_event` is a multiprocessing event, nevermind

untitaker (approved):

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -695,6 +695,7 @@ def shutdown(self) -> None:
         Shutdown cleanly
         Activate the shutdown event and drain results before terminating children.
         """
+        CHILD_JOIN_TIMEOUT_SEC = 5
```
markstory: Should this be at the module level? Or defined in the constants module?

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -160,7 +201,41 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
+                # we just care if result() raises or not
+                [f.result(RESULT_TIMEOUT_SEC) for f in task.pending_futures]
+            # If any pending producer futures failed, retry the task
+            except Exception:
+                task.status = TASK_ACTIVATION_STATUS_FAILURE
+                task_func = _get_known_task(task.inflight.activation)
+                if task_func and task_func.retry:
+                    retry_state = task.inflight.activation.retry_state
+                    if not task_func.retry.max_attempts_reached(retry_state):
+                        task.status = TASK_ACTIVATION_STATUS_RETRY
```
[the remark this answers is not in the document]
bmckerry: This is intentional, future resolution should be hidden from the user IMO, they shouldn't have to add it to their tasks retry policy.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -706,7 +707,10 @@ def shutdown(self) -> None:
         for child in self._children:
             child.terminate()
         for child in self._children:
-            child.join()
+            child.join(CHILD_JOIN_TIMEOUT_SEC)
+            if child.is_alive():
+                child.kill()
+                child.join()
```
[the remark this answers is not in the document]
bmckerry: If this happens, I believe those tasks will be picked up and retried at a later point.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -160,7 +201,41 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
+                next_state=task.status,
+                execution_start_time=task.execution_start_time,
+                futures_start_time=task.futures_start_time,
+            )
+
+        def check_task_future_completion() -> None:
+            if len(pending_task_futures) > 0:
+                for task in pending_task_futures.copy():
+                    if all([f.done() for f in task.pending_futures]):
+                        await_task_futures(task)
```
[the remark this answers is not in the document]
bmckerry: Not sure when this would happen for producer futures, but if it does then the task will get retried after the worker child is killed.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -213,7 +254,41 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
+            """
+            Blocks on the result of each producer futures in the given task.
+            If any futures raise an exception, the task is considered failed
+            and submitted for retry (if the policy allows).
+            """
+            RESULT_TIMEOUT_SEC = 1
+            try:
+                # We don't care about the actual result value,
+                # we just care if result() raises or not
+                [f.result(RESULT_TIMEOUT_SEC) for f in task.pending_futures]
```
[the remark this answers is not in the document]
bmckerry: This is untrue

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -213,7 +254,40 @@ def handle_alarm(signum: int, frame: FrameType | None) -> None:
...
+                inflight=task.inflight,
+                next_state=task.status,
+                execution_start_time=task.execution_start_time,
+                futures_start_time=task.futures_start_time,
+            )
+
+        def check_task_future_completion() -> None:
+            if len(pending_task_futures) > 0:
+                for task in pending_task_futures.copy():
+                    if all([f.done() for f in task.pending_futures]):
```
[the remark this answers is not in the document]
bmckerry: https://github.com/getsentry/taskbroker/pull/642#discussion_r3276543821

clients/python/src/taskbroker_client/worker/workerchild.py
fpacifici: Not sure I follow the rationale. If the parent process is already managing the shutdown triggering the shutdown on the children why do we need to intercept sigterm here ? 
It should be up to the main process to tell the children when to shut down. That would make the process a lot easier to follow and more deterministic.

bmckerry: > It should be up to the main process to tell the children when to shut down

I agree, my understanding is that when the main process receives SIGINT it propagates the signal to all child processes as well (if I'm understanding [this thread](https://stackoverflow.com/questions/17704093/python-signals-for-multiprocessing) correctly). By ignoring the signal in child processes, we allow the main process to shut down all child processes when getting a SIGINT (via the standard shutdown process).

I'm not sure when SIGINT is sent other than when using `ctrl + c` in a dev environment, so this probably isn't a huge deal and I can remove this if you'd prefer.

markstory: > I'm not sure when SIGINT is sent other than when using ctrl + c in a dev environment, so this probably isn't a huge deal and I can remove this if you'd prefer.

I think that is the main scenario it gets used. K8s uses SIGTERM.

clients/python/src/taskbroker_client/worker/worker.py
bmckerry: Didn't notice `constants.py`, I'll move there. Thanks

PR #645, opened 2026-05-27
ref: Drop "Inflight" prefix from activation types and modules

evanh:
## Summary
- Rename Rust types `InflightActivation`, `InflightActivationStore` (trait), `InflightActivationStatus`, `InflightActivationBuilder`, `InflightActivationStoreConfig`, `InflightActivationBatcher`, `InflightActivationWriter` to their unprefixed equivalents.
- Rename `src/kafka/inflight_activation_{batcher,writer}.rs` to `activation_{batcher,writer}.rs`.
- Update doc comments, the AGENTS.md style example, and `bench_InflightActivationStore` → `bench_ActivationStore`.

The `Inflight` prefix wasn't actually disambiguating against anything — the only nearby type is the protobuf `TaskActivation`, and the `TaskActivation` vs `Activation` contrast reads more clearly without the extra modifier.

## Intentionally not changed
- The `inflight_taskactivations` SQL table and migration filenames (renaming would require a new migration and is out of scope).
- The Python `InflightTaskActivation` wrapper — `Inflight` had to stay because dropping just it would collide with the imported protobuf `TaskActivation`.
- Metric names (`consumer.inflight_activation_writer.*`) — kept the old names to avoid breaking dashboards/alerts.
- Local variable names like `inflight`, `inflight_opt`, `inflight_msgs`, and the `taskbroker-inflight.sqlite` default path.
- Persisted DB status strings (unchanged because `Display`/`FromStr` for `ActivationStatus` use the variant names, which weren't renamed).

## Test plan
- [x] `cargo check --all-targets`
- [x] `cargo build --release`
- [x] `cargo test --lib` (246 passed, 0 failed)
- [x] `make format`
- [x] `make lint`
- [x] Integration tests in CI

### DIFF
--- a/benches/store_bench.rs
+++ b/benches/store_bench.rs
@@ -5,9 +5,9 @@ use criterion::{Criterion, criterion_group, criterion_main};
 use rand::Rng;
 use tokio::task::JoinSet;
 
-use taskbroker::store::activation::InflightActivationStatus;
-use taskbroker::store::adapters::sqlite::{InflightActivationStoreConfig, SqliteActivationStore};
-use taskbroker::store::traits::InflightActivationStore;
+use taskbroker::store::activation::ActivationStatus;
+use taskbroker::store::adapters::sqlite::{ActivationStoreConfig, SqliteActivationStore};
+use taskbroker::store::traits::ActivationStore;
 use taskbroker::test_utils::{
     generate_temp_filename, generate_unique_namespace, make_activations_with_namespace,
 };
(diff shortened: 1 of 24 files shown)


untitaker (approved):

benches/store_bench.rs
```diff
@@ -89,9 +89,9 @@ async fn set_status(num_activations: u32, num_workers: u32) {
     };
 
     let store = Arc::new(
-        SqliteActivationStore::new(
+        SqliteStore::new(
```
untitaker, line 92: this rename didn't seem necessary. this store indeed is for activations

markstory (approved):

PR #654, opened 2026-05-29
feat(taskbroker): Batch Claimed to Processing Updates

george-sentry:
## Linear
Refs [STREAM-920](https://linear.app/getsentry/issue/STREAM-920/batch-claim-processing-updates)

## Description
After a "claimed" activation is pushed successfully, it needs to be updated to "processing." Right now this happens one at a time, which is a performance bottleneck. This PR adds optional batching for these updates, which adds several thousand additional tasks per second on my typical testing workload (100 millisecond dummy tasks).

Concerning data loss, what if taskbroker crashes while these updates are queued? Those updates are lost, meaning even though those tasks were claimed and sent successfully, they weren't updated to processing. Three things can happen here...

- If the worker completes a task fast enough, it will be updated to "completed" directly from "claimed" and later removed by upkeep ✅
- If the worker doesn't complete a task fast enough, it may be reset back to "pending" before the worker completes it. In that case, it will be updated from "pending" directly to "completed" and later removed by upkeep ✅
- If the worker doesn't complete a task fast enough and the taskbroker has enough time to reclaim and / or resend it again, the task will be completed twice ✅ ✅

In all cases, tasks aren't dropped but may be repeated.


src/push/updater.rs
```diff
@@ -0,0 +1,174 @@
...
+                    break;
+                }
+
+                _ = interval.tick(), if self.config.batch_push_updates => {
+                    // Lock the ID buffer
+                    let mut buffer = self.buffer.lock().await;
+
+                    // Make sure we aren't flushing too soon
+                    let now = Utc::now().timestamp_millis();
+                    let elapsed = self.last_flush.timestamp_millis()  - now;
```
markstory: Shouldn't this be `now - last_update`?

src/push/mod.rs
```diff
@@ -53,9 +55,15 @@ impl PushPool {
...
     #[framed]
-    pub async fn start(&self, workers: Vec<WorkerMap>) -> Result<()> {
+    pub async fn start(&self, workers: Vec<WorkerMap>, updater: Arc<dyn Updater>) -> Result<()> {
         let mut workers = workers.into_iter();
 
+        // Start the updater
+        let updaterd = tokio::spawn({
+            let updater = updater.clone();
+            async move { updater.start().await }
+        });
```
markstory: How do the push threads append to this updater?

george-sentry: I forgot to pass it into the push threads. That's fixed now.

src/push/updater.rs
george-sentry: Yepp, thanks!

src/push/updater.rs
```diff
@@ -0,0 +1,221 @@
...
+
+        loop {
+            tokio::select! {
+                _ = self.stop.notified() => {
+                    info!("Stopping lazy updater...");
+                    break;
+                }
+
+                _ = interval.tick(), if self.config.batch_push_updates => {
+                    // Lock the ID buffer
```
evanh, line 145: There is a potential bottleneck here: The buffer is locked for the entire lifetime of the flush operation. Really it only needs to be locked while it is drained, and then again if/when there is an error and the IDs need to be added back to the  buffer. This lock is being held over the DB update as well. I might move this lock into the `flush` method so that the buffer can be locked/unlocked just for the `drain` call and the `extend` call.

src/push/updater.rs
```diff
@@ -0,0 +1,221 @@
...
+                        error = ?e,
+                        "Failed to flush claimed → processing updates (full buffer)"
+                    );
+
+                    return Err(e);
+                }
+            }
+        }
+
+        buffer.push(id);
```
evanh, line 220: Why do we not push the ID first, and then flush?

src/push/updater.rs
evanh: Note that this can be done simply:

```
let ids = {
       let buffer = self.buffer.lock().await;
       buffer.drain.collect()
};
```

evanh: Note too that if the lock is managed in this way, we don't need a tokio `Mutex` and can use a `std` mutex.

src/push/updater.rs
george-sentry: In that case, the buffer would grow infinitely when DB is slow or unresponsive causing it to run out of memory.

george-sentry: By flushing first, we guarantee that the buffer never grows above the limit.

src/push/updater.rs
george-sentry: I'm worried that this may cause the buffer to grow without bounds. Suppose you drain the buffer but fail to perform the update. When you put those IDs back into the buffer, it will have accumulated a significant number of IDs in the meantime since it's unlocked. Now your buffer is larger than the configured maximum. If this keeps happening, it will grow larger and larger. By keeping the buffer locked while flushing, we prevent any more tasks from being appended until the flush is complete.

george-sentry: Also, my channel implementation of batching works in a similar manner with excellent performance, because while it's flushing, it stops receiving messages from the channel. The channel is bounded, so eventually it fills up and the push thread must wait until it can submit an ID (equivalent to waiting for the lock before submitting an ID). In other words, I don't think this is a bottleneck.

evanh: OK. Can you please make sure to have a metric that will expose how much contention exists on the buffer lock?

evanh (approved):
Approved, assuming the metric I mentioned exists.

src/push/updater.rs
george-sentry: Done!

PR #659, opened 2026-05-29
feat: Add default metrics tags to Sentry errors

untitaker:
## Summary
- Reuse the default metrics tags (e.g. pool name, latency_tier) as Sentry tags so errors can be filtered/correlated with specific taskbroker instances
- Add a `taskbroker::tokio::spawn` helper that propagates the main Sentry hub to spawned tasks, ensuring tags are available across all threads

## Test plan
- [ ] Deploy and verify tags appear on Sentry errors

### DIFF
--- a/src/tokio.rs
+++ b/src/tokio.rs
@@ -1,5 +1,20 @@
 use std::future::Future;
-use tokio::task::JoinSet;
+
+use sentry::{Hub, SentryFutureExt};
+use tokio::task::{JoinHandle, JoinSet};
+
+/// Spawns a task with the main Sentry hub propagated.
+///
+/// This ensures that global Sentry tags configured via `configure_scope` on
+/// the main thread are available in spawned tasks. Without this, tokio worker
+/// threads would have their own hub without those tags.
+pub fn spawn<F>(future: F) -> JoinHandle<F::Output>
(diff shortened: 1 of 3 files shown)


src/tokio.rs
```diff
@@ -1,5 +1,20 @@
...
+///
+/// This ensures that global Sentry tags configured via `configure_scope` on
+/// the main thread are available in spawned tasks. Without this, tokio worker
+/// threads would have their own hub without those tags.
+pub fn spawn<F>(future: F) -> JoinHandle<F::Output>
+where
+    F: Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    tokio::spawn(future.bind_hub(Hub::new_from_top(Hub::main())))
```
george-sentry, line 16: The compiler will probably inline all of this because it's so simple, but what if this was a macro?

untitaker: do you want this to be a macro in order for it to inline better? not sure I understand

george-sentry (approved):

src/tokio.rs
george-sentry: Ehh no that's okay. Unnecessary complexity.

PR #664, opened 2026-06-02
feat(config): Add Validation to Several Config Fields

george-sentry:
## Linear
Refs [STREAM-843](https://linear.app/getsentry/issue/STREAM-843/better-configuration-organization)

## Description
Some configuration options have specific requirements, like they must be greater than zero or a power of two. Until now, we've relied on ad hoc assertions and calls to functions like `max(1)` to enforce these constraints. I think it would be better to guarantee these constraints immediately on startup because ad hoc checks can easily be forgotten.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -2,12 +2,15 @@
 use std::borrow::Cow;
 use std::collections::BTreeMap;
 
+use anyhow::Result;
 use figment::providers::{Env, Format, Yaml};
 use figment::{Figment, Metadata, Profile, Provider};
 use rdkafka::ClientConfig;
 use serde::{Deserialize, Serialize};
+use validator::{Validate, ValidationError};
 
 use crate::Args;
+use crate::fetch::MAX_FETCH_THREADS;
 use crate::logging::LogFormat;
 
 #[derive(Clone, Copy, Debug, Eq, PartialEq, Hash, Deserialize, Serialize)]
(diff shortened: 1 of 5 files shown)


untitaker (approved):
validation annotations are good, please make sure the design is still consistent with https://github.com/getsentry/taskbroker/pull/663 where we have another validation method on `Config`

untitaker:
one thing to think about: `impl Validate for Config` inherently exposes a public validation function. is there a way to make it private (and call it in `validate_and_normalize`), or maybe `validate_and_normalize` can be made private and called via this validate crate?

https://github.com/Keats/validator/blob/master/README.md#struct-level-validation

src/config.rs
```diff
@@ -570,16 +580,23 @@ impl Default for Config {
...
         if let Some(path) = &args.config {
             builder = builder.merge(Yaml::file(path));
         }
-        // Use split("__") to support nested config via envvars like:
-        // TASKBROKER_KAFKA_TOPICS__PROFILES__CLUSTER=my-cluster
+
+        // Use "__" for nested configurations via environment variables, like `TASKBROKER_KAFKA_TOPICS__PROFILES__CLUSTER`
         builder = builder.merge(Env::prefixed("TASKBROKER_").split("__"));
         let mut config: Config = builder.extract()?;
+
```
untitaker, line 593: @george-sentry I think we should not have two of these functions. `normalize_and_validate` is called from tests, and in those tests we should call `validate` too IMO

PR #667, opened 2026-06-02
doc: Add docs for config migration

untitaker:
https://github.com/getsentry/taskbroker/pull/663

ref STREAM-1042

This is probably useful for self-hosted.


evanh (approved):
Thanks for this!

PR #671, opened 2026-06-03
ref(workerchild): add metric tracking pending futures

bmckerry:
Adds a metric to get a picture of how many activations with pending futures a processing pool has.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/workerchild.py
+++ b/clients/python/src/taskbroker_client/worker/workerchild.py
@@ -284,17 +284,16 @@ def child_process(
             )
 
         def check_task_future_completion() -> None:
+            # Records how many activations with pending producer futures
+            # the worker child has.
+            metrics.gauge(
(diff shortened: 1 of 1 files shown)


evanh:
> I was going to make this a gauge

You could log both? One would tell you the total/p99, and the other would tell you how many produce futures a single workerchild has to deal with.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -284,6 +284,16 @@ def await_task_futures(task: ActivationWithPendingFutures) -> None:
             )
 
         def check_task_future_completion() -> None:
+            # Records a distribution of how many activations with pending producer futures
+            # a processing pool has across all worker child processes.
+            # Only reports if the worker actually has pending futures.
+            metrics.distribution(
```
untitaker: i suspect this should be a gauge?

untitaker (approved):
approving but one question

PR #672, opened 2026-06-03
fix(client): log push-mode request deserializer failures

untitaker:
## Summary

In **push mode** the Python worker runs a gRPC server (`PushTaskWorker` → `WorkerServicer`). When a `PushTask` request body fails to deserialize — proto schema mismatch between broker and worker, or a corrupted payload — the underlying exception is **silently swallowed** and the broker only sees an opaque `INTERNAL` status. Nothing lands on our own logger and no metric is emitted, so the failure is effectively invisible in worker logs.

This PR logs the exception on the `taskbroker_client.worker.client` logger before re-raising.

## Where the exception is swallowed

The request deserializer (our `RequestSignatureServerInterceptor` wraps it) is invoked by gRPC, and gRPC catches *any* exception it raises. Links below are pinned to the `grpcio==1.75.1` tag (the version this repo locks).

1. **The swallow** — [`grpc/_common.py` `_transform()`, lines 88–92](https://github.com/grpc/grpc/blob/v1.75.1/src/python/grpcio/grpc/_common.py#L88-L92):
   ```python
   try:
       return transformer(message)        # <- our deserializer
   except Exception:                      # pylint: disable=broad-except
       _LOGGER.exception(exception_message)   # logged on the `grpc._common` logger only
       return None                            # exception turned into None
   ```
   [`deserialize()`](https://github.com/grpc/grpc/blob/v1.75.1/src/python/grpcio/grpc/_common.py#L99-L104) calls this with `exception_message="Exception deserializing message!"`.

2. **The opaque status** — [`grpc/_server.py` `_receive_message()`, lines 333–343](https://github.com/grpc/grpc/blob/v1.75.1/src/python/grpcio/grpc/_server.py#L333-L343):
   ```python
   request = _common.deserialize(serialized_request, request_deserializer)
   ...
   if request is None:
       _abort(state, call, cygrpc.StatusCode.internal,
              b"Exception deserializing request!")
   ```

So the real traceback exists only on the `grpc._common` logger at `ERROR`. Unless the worker explicitly routes gRPC's internal loggers to a handler, it is lost — and the broker just gets `INTERNAL "Exception deserializing request!"` with no detail about *why*.

## Fix

Wrap the deserializer body in the interceptor and log on our own logger before re-raising (gRPC still produces the `INTERNAL` status for the caller, but now there's a visible, attributable record on the worker side):

```python
try:
    return inner_deserializer(serialized_request)
except Exception:
    logger.exception(
        "taskworker.grpc_server.request_deserialization_failed",
        extra={"method": method},
    )
    raise
```

## Test

`test_request_signature_server_interceptor_logs_deserialization_errors` stands up a real `grpc.server` with the interceptor, sends a correctly-signed but unparseable body, and asserts:
- the caller receives `StatusCode.INTERNAL`
- the servicer never runs
- a `taskworker.grpc_server.request_deserialization_failed` record is emitted on our logger

Verified the test **fails without the fix** (no such log record) and **passes with it**. A real server is required because the deserializer runs on gRPC's event-polling thread, not the servicer thread — a mocked channel cannot exercise this path.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/client.py
+++ b/clients/python/src/taskbroker_client/worker/client.py
@@ -225,7 +225,18 @@ class RequestSignatureServerInterceptor(ServerInterceptor):
                 _RPC_SIGNATURE_AUTH_TLS.failed = True
                 return inner_deserializer(b"")
 
-            return inner_deserializer(serialized_request)
+            try:
+                return inner_deserializer(serialized_request)
+            except Exception:
+                # gRPC swallows deserializer exceptions in `grpc._common._transform`
+                # (it logs to the `grpc._common` logger and returns None), and the server
+                # then aborts the call with an opaque INTERNAL "Exception deserializing
+                # request!". Log here so the failure is visible on our own logger.
+                logger.exception(
+                    "taskworker.grpc_server.request_deserialization_failed",
+                    extra={"method": method},
+                )
+                raise
 
         def unary_unary(request: Any, context: grpc.ServicerContext) -> Any:
             if getattr(_RPC_SIGNATURE_AUTH_TLS, "failed", False):
--- a/clients/python/tests/worker/test_client.py
+++ b/clients/python/tests/worker/test_client.py
@@ -5,7 +5,9 @@ import random
 import string
 import time
 from collections import defaultdict
-from collections.abc import Callable
+from collections.abc import Callable, Generator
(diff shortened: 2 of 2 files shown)


untitaker:
bogus actually. i'll just check the grpc logger setup in sentry.

PR #680, opened 2026-06-05
fix: Update latency check to use CLAIMED status

evanh:
There was a chance that the latency metric would be very spiky if a message was moving between the PENDING and CLAIMED status. Update the check to use both statuses so the latency stays consistent.

On a related note, update the max message size default. The largest Kafka topic size is currently 50MB, but we actually want a limit slightly higher than that, since for raw taskbrokers we wrap the Kafka message in a few more bytes.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -609,7 +609,7 @@ impl Default for Config {
             maintenance_task_interval_ms: 6000,
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 5000000,
-            grpc_max_message_size: 10 * 1024 * 1024, // 10MB
+            grpc_max_message_size: 12 * 1024 * 1024, // 12MB
             vacuum_page_count: None,
             full_vacuum_on_start: true,
             full_vacuum_on_upkeep: true,
--- a/src/config.rs
+++ b/src/config.rs
@@ -609,7 +609,7 @@ impl Default for Config {
             maintenance_task_interval_ms: 6000,
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 5000000,
-            grpc_max_message_size: 12 * 1024 * 1024, // 12MB
+            grpc_max_message_size: 52 * 1024 * 1024, // 52MB
             vacuum_page_count: None,
             full_vacuum_on_start: true,
             full_vacuum_on_upkeep: true,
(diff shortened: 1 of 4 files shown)


src/store/tests.rs
```diff
@@ -1910,10 +1910,12 @@ async fn test_pending_activation_max_lag_use_oldest(#[case] adapter: &str) {
...
-    let mut pending = make_activations(2);
-    pending[0].received_at = now - Duration::from_secs(10);
-    pending[1].received_at = now - Duration::from_secs(500);
-    assert!(store.store(pending).await.is_ok());
+    let mut activations = make_activations(3);
+    activations[0].received_at = now - Duration::from_secs(10);
+    activations[1].received_at = now - Duration::from_secs(500);
+    activations[2].status = ActivationStatus::Claimed;
+    activations[2].received_at = now - Duration::from_secs(50);
+    assert!(store.store(activations).await.is_ok());
```
markstory, line 1940: Should there also be a test for the claimed task having the highest latency?

src/config.rs
```diff
@@ -609,7 +609,7 @@ impl Default for Config {
             maintenance_task_interval_ms: 6000,
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 5000000,
-            grpc_max_message_size: 10 * 1024 * 1024, // 10MB
+            grpc_max_message_size: 12 * 1024 * 1024, // 12MB
```
markstory: Was this on purpose?

evanh: Yeah, updated description.

untitaker (approved):
> On a related note, update the max message size default. The default Kafka topic size is 10MB, but we actually want a limit slightly higher than that, since for raw taskbrokers we wrap the Kafka message in a few more bytes.

there is an infra PR that sets it to 50 MB for profiles. i suggest to make 50 MB the default in code and remove any defaults from infra, so we don't have to worry about this anymore. (otherwise we'll have this problem with self-hosted again)

markstory:
>  i suggest to make 50 MB the default in code and remove any defaults from infra, so we don't have to worry about this anymore. (otherwise we'll have this problem with self-hosted again)

50MB payloads will be rough on SQLite and postgres because of the write throughput. Should we also be applying compression to the large profile payloads?

untitaker:
> Should we also be applying compression to the large profile payloads?

I can follow up on this. fwiw 50 MB is not avg, but max.

evanh:
> > i suggest to make 50 MB the default in code and remove any defaults from infra, so we don't have to worry about this anymore. (otherwise we'll have this problem with self-hosted again)
> 
> 50MB payloads will be rough on SQLite and postgres because of the write throughput. Should we also be applying compression to the large profile payloads?

There is already a ticket to introduce that: https://linear.app/getsentry/issue/STREAM-1107/raw-mode-needs-to-compress-the-payload

PR #683, opened 2026-06-08
fix(worker): keep worker health checks alive while idle in push mode

lvthanh03:
Refs STREAM-1056

Fixes push-mode worker health checks so low-throughput or idle workers continue touching the health check file.

Previously, push mode only emitted the health check from `update_task`, which only runs after a task result is sent. Unlike pull mode, idle push workers had no predictable heartbeat and could be killed despite being healthy.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/client.py
+++ b/clients/python/src/taskbroker_client/worker/client.py
@@ -560,6 +560,9 @@ class PushTaskbrokerClient:
             channel = grpc.intercept_channel(channel, RequestSignatureInterceptor(secrets))
         return ConsumerServiceStub(channel)
 
+    def emit_health_check(self) -> None:
+        self._emit_health_check()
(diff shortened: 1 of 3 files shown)


evanh (approved):
Nice!

george-sentry (approved):
We should do a quick test to make sure this doesn't significantly impact performance, but otherwise looks good!

lvthanh03:
> We should do a quick test to make sure this doesn't significantly impact performance, but otherwise looks good!

Tested on 1 broker, 22 workers (32 processes/worker) and example task which takes 100ms, we are processsing 6.9k tasks/s on a theoretical limit of 7,040 tasks/s, so on this configuration there is no impact of the healthcheck thread on the workers.

The healthcheck file now also updates even when there are no tasks / worker is idle.

PR #706, opened 2026-06-15
feat(config): (12) Change `query_max_retries` from `Option<u32>` to `u32`

george-sentry:
## Linear
Refs [STREAM-843](https://linear.app/getsentry/issue/STREAM-843/better-configuration-organization)

## Description
This is **PART 12** of my configuration improvement effort. I am now trying to get rid of the `RetryConfig` in `store/retry.rs`, create a new `RetryConfig` in `config/store.rs`, and nest it inside of `StoreConfig`.

I am doing this so that I can pass `StoreConfig` directly to `PostgresStore` and `SqliteStore` instead of constructing intermediate `PostgresStoreConfig` and `SqliteStoreConfig` structs.

In this PR, I change `db_query_max_retries` to take a `u32` with a default of zero instead of an `Option<u32>` because when we derive `RetryConfig` in `store/retry.rs`, we default to zero when it's `None`. So when it's not provided, we should use 0 for that value. Why not make that more explicit?

Furthermore, since it's an `Option`, you would expect the default to be `None`. Yet it's `Some(3)`. That's confusing.

So let's just make it a simple `u32`, and when it's not provided, that means the user doesn't want to retry queries, so we should just set it to zero. Easy!


src/config/store.rs
```diff
@@ -179,7 +179,7 @@ impl Default for StoreConfig {
             pg: PgConfig::default(),
             sqlite: SqliteConfig::default(),
             db_write_failure_backoff_ms: 4000,
-            db_query_max_retries: Some(3),
+            db_query_max_retries: 0, // Do not retry by default
```
untitaker: this is a behavior change and i have no idea if this is what we want. you probably know better.

we should expect that self-hosted and maybe single-tenant will not retry queries. are we ok w that?

george-sentry: Ehh good point. I'm not sure actually. Vic would know better because she implemented retries, but she's out for the week. Would you agree that if I change this to `3` by default, the behavior will be the same as it was before when `db_query_max_retries` was an `Option`?

untitaker: if this defaults to 3 then i guess it's not a behavior change and there's no issue

untitaker (approved):

src/config/store.rs
```diff
@@ -120,8 +122,8 @@ pub struct StoreConfig {
     pub db_write_failure_backoff_ms: u64,
 
     /// The maximum number of times to retry a transient database query error
-    /// before surfacing the error. When None, queries are not retried.
-    pub db_query_max_retries: Option<u32>,
+    /// before surfacing the error. When zero, queries are not retried.
```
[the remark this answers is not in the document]
george-sentry: Nobody uses `store.db_query_max_retries` yet so that's OK.

PR #714, opened 2026-06-17
fix: Only need to skip batching for actual RETRY status

evanh:


untitaker (approved):

enochtangg (approved):

untitaker (commented):
Note, this can be simplified like that, but the reason we can't batch those status updates is specifically because of these two fields. Retry updates don't inherently have that problem, and not all retry updates have those two fields.

PR #719, opened 2026-06-17
ref(worker): add more observability for producer futures

bmckerry:
This PR adds more observability to TaskProducer (and the worker child) to help troubleshoot why some producer futures never got marked as done.

Specifically:
- changes TaskProducer's `_pending_futures` from a single queue to a dict keyed by producer name, so we can break down worker child metrics by producer
- adds metric for how many messages were produced by a task, tagged by producer name
- separates the task execution time metric from the time a task that produced spends sitting in the worker child's pending queue
  - adds a new `e2e` metric that records both of these put together
- records a gauge of how many futures are still incomplete in a worker child's tasks with pending futures
  - not sure if this should be a dist?
- records a gauge of how long the oldest task with pending futures is
  - previously only tasks with 100% complete futures had their latency recorded, this will tell us how old the tasks with incomplete futures are

### DIFF
--- a/clients/python/src/taskbroker_client/worker/workerchild.py
+++ b/clients/python/src/taskbroker_client/worker/workerchild.py
@@ -324,7 +324,7 @@ def child_process(
                         await_task_futures(task)
                     else:
                         # How many futures are still pending in this task
-                        metrics.gauge(
+                        metrics.distribution(
                             "taskworker.task.incomplete_futures",
                             len([f for f in future_status if not f]),
                             tags={
@@ -335,7 +335,7 @@ def child_process(
                         )
                 # How long has the oldest pending task been sitting in the queue
                 if oldest := get_oldest_pending_activation():
-                    metrics.gauge(
+                    metrics.distribution(
                         "taskworker.worker.oldest_pending_activation_age",
                         time.time() - oldest.futures_start_time,
                         tags={
--- a/clients/python/src/taskbroker_client/worker/workerchild.py
+++ b/clients/python/src/taskbroker_client/worker/workerchild.py
@@ -66,7 +66,7 @@ class ActivationWithPendingFutures:
     status: TaskActivationStatus.ValueType
     execution_start_time: float
     futures_start_time: float
-    pending_futures: set[ProducerFuture[BrokerValue[KafkaPayload]]]
+    pending_futures: dict[str, set[ProducerFuture[BrokerValue[KafkaPayload]]]]
     task_func: Task[Any, Any]
 
 
@@ -263,24 +263,27 @@ def child_process(
             and submitted for retry (if the policy allows).
             """
             RESULT_TIMEOUT_SEC = 1
+            error_name: str = "N/A"
             try:
                 # We don't care about the actual result value,
                 # we just care if result() raises or not
-                [f.result(RESULT_TIMEOUT_SEC) for f in task.pending_futures]
+                [f.result(RESULT_TIMEOUT_SEC) for fut in task.pending_futures.values() for f in fut]
                 produce_status = "success"
             # If any pending producer futures failed, retry the task
-            except Exception:
+            except Exception as e:
                 task.status = TASK_ACTIVATION_STATUS_FAILURE
                 if task.task_func.retry:
                     retry_state = task.inflight.activation.retry_state
                     if not task.task_func.retry.max_attempts_reached(retry_state):
                         task.status = TASK_ACTIVATION_STATUS_RETRY
                 produce_status = "failure"
+                error_name = type(e).__name__
             metrics.incr(
                 "taskworker.worker.produce.result",
                 tags={
                     "status": produce_status,
                     "processing_pool": processing_pool_name,
+                    "error_name": error_name,
                 },
             )
             pending_task_futures.remove(task)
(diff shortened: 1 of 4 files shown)


clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -636,6 +686,18 @@ def record_task_execution(
...
+        metrics.distribution(
+            "taskworker.worker.e2e_latency",
+            (execution_duration + futures_duration),
+            tags={
+                "namespace": activation.namespace,
+                "taskname": activation.taskname,
+                "processing_pool": processing_pool_name,
+                "taskbroker_host": taskbroker_host,
+            },
+        )
```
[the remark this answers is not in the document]
bmckerry: I'm not sure how we'd want to handle these - in theory tasks aren't "complete" until all produced futures are complete, but I still see task function execution time & future awaiting time as separate things we should track

Open to advice here

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -263,35 +263,50 @@ def await_task_futures(task: ActivationWithPendingFutures) -> None:
...
                     "processing_pool": processing_pool_name,
+                    "error_name": error_name,
                 },
             )
             pending_task_futures.remove(task)
             _task_execution_complete(
                 inflight=task.inflight,
                 next_state=task.status,
                 execution_start_time=task.execution_start_time,
+                execution_end_time=task.futures_start_time,
```
untitaker, line 294: looks like you're already passing in `futures_start_time`

clients/python/src/taskbroker_client/worker/workerchild.py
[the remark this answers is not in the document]
untitaker: it seems to me that what we're measuring here is the additional time it takes to move futures from one list to another, not waiting for the produce to actually finished. is that correct?

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -304,8 +319,31 @@ def check_task_future_completion() -> None:
                     },
                 )
                 for task in pending_task_futures.copy():
-                    if all([f.done() for f in task.pending_futures]):
+                    future_status = [f.done() for fut in task.pending_futures.values() for f in fut]
+                    if all(future_status):
                         await_task_futures(task)
+                    else:
+                        # How many futures are still pending in this task
+                        metrics.gauge(
```
markstory: I think you might want a histogram here instead. Each worker child will observe different gauge values, and your metric doesn't have enough dimensions to separate all the possible values.

clients/python/src/taskbroker_client/worker/workerchild.py
bmckerry: In this branch of the code, tasks have been sitting in `pending_task_futures` waiting for their producer futures to finish (so `execution_end_time` == `futures_start_time`, which is when the tasks were added to `pending_task_futures`).

[Here](https://github.com/getsentry/taskbroker/pull/719/changes/BASE..c6f52d0846cce786e4a3ac6419a4304dd19a09f4#diff-5887a149071005b81272996d7424a5d31c124507cd28041827444c0af3ffa7c4R511) is where we call `_task_execution_complete()` when a task had no pending futures, so a different value (the current time) is passed in as `execution_end_time`.

I think this module is due for a bit of a refactor to clean up the branching logic, but I'd rather do that in a separate PR.

clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -304,8 +319,31 @@ def check_task_future_completion() -> None:
...
+                            len([f for f in future_status if not f]),
+                            tags={
+                                "processing_pool": processing_pool_name,
+                                "namespace": task.inflight.activation.namespace,
+                                "taskname": task.inflight.activation.taskname,
+                            },
+                        )
+                # How long has the oldest pending task been sitting in the queue
+                if oldest := get_oldest_pending_activation():
+                    metrics.gauge(
```
markstory: Same here for histogram over gauge.

clients/python/src/taskbroker_client/worker/workerchild.py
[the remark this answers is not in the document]
bmckerry: `execution_duration` is how long the task function takes to execute, `futures_duration` is how long it took for all of a tasks producer futures to be marked as done. Since tasks that produced messages don't go into `processed_tasks` until all their futures are done, this e2e time metric tracks how long that took.

clients/python/src/taskbroker_client/worker/workerchild.py
bmckerry: This is something I'm still confused on - in https://github.com/getsentry/taskbroker/pull/671 I added a metric tracking a value that's separate for each worker child and was told it should be a gauge rather than a dist. @fpacifici raised the point that with how we handle distribution metrics in Sentry atm, basically all values except `max` and `avg` are wrong and we should just use gauges. I'm still not sure what's the right approach for situations like this.

markstory: Lets try a gauge then :smile:

untitaker: i agree with mark here, unless we're going to add tags for each taskworker subprocess (maybe this is something we can do with sentry metrics where every datapoint costs the same), a gauge with too few tags is worse than putting the same gauge into a histogram, regardless of whether semantically it's supposed to be a gauge

untitaker (approved):
approving to unblock

i think the histogram vs gauge thing might prevent you from getting insights. but if you feel like it's good enough for your purposes to see workerchild's metrics muddled together per-pool (and I think it'll be fine), then i'd rather unblock for now and revisit later (i can see this discussion dragging on for too long, especially since it concerns cardinality)

and the other thing about which latency is the right one: i also don't think we need to resolve this conversation in this PR

it's important to me that we get to a state where we can safely roll out producer futures in existing high-scale tasks (and also for tasks that we will create from legacy consumers), and for that we need the investigation to continue. otherwise legacy consumers project will soon get to a point where we can't port more consumers either. everything else is irrelevant right now.

PR #731, opened 2026-06-25
feat(worker): defer SERVING until child processes are warm

enochtangg:
Refs: [STREAM-1244](https://linear.app/getsentry/issue/STREAM-1244)

In push mode, a new taskworker pod advertises gRPC health `SERVING` immediately after `server.start()`, with zero warm children. Each child must run `import_app()` + `app.load_modules()` (~30-40s) before it consumes from the queue. Because the k8s readiness probe and the GCP NEG both trust this gRPC health signal, the pod enters Envoy's routable set ~30s before it can actually serve. As a result, the child queue fills, the worker returns busy, and the broker's bounded push pool jams on the slow pushes.

This PR is responsible for holding the gRPC health service at `NOT_SERVING` until children are warm. Changes:
- the child_process takes an optional `ready_counter` (`multiprocessing.Value`) and increments it once, right before the consume loop. The increment happens after `load_modules()` and after the SIGTERM/SIGINT handlers are installed, so any child counted as warm can also drain on shutdown.
- Sets`ready_count` in parent thread  and passes it to each child.
- On startup, updated the push worker to  `_await_children_warm()` between `server.start()` and the `SERVING` flip. Polling  `ready_count >= min_ready` every 250ms tick. This is also bounded by `warmup_timeout`.

Testing:
- Tested changes locally
- Added unit tests

### DIFF
--- a/clients/python/src/taskbroker_client/constants.py
+++ b/clients/python/src/taskbroker_client/constants.py
@@ -57,6 +57,12 @@ DEFAULT_WORKER_HEALTH_CHECK_SEC_PER_TOUCH = 1.0
 The number of gRPC requests before touching the health check file
 """
 
+DEFAULT_WORKER_WARMUP_TIMEOUT_SEC = 90.0
+"""
+Max seconds PushTaskWorker waits for >= min_ready children to warm up
+before flipping gRPC health to SERVING anyway.
+"""
+
 
 ALWAYS_EAGER = False
 """
--- a/clients/python/src/taskbroker_client/constants.py
+++ b/clients/python/src/taskbroker_client/constants.py
@@ -59,8 +59,8 @@ The number of gRPC requests before touching the health check file
 
 DEFAULT_WORKER_WARMUP_TIMEOUT_SEC = 90.0
 """
-Max seconds PushTaskWorker waits for >= min_ready children to warm up
-before flipping gRPC health to SERVING anyway.
+Max seconds PushTaskWorker waits for all children to warm up before
+flipping gRPC health to SERVING anyway.
 """
 
 
--- a/clients/python/src/taskbroker_client/worker/worker.py
+++ b/clients/python/src/taskbroker_client/worker/worker.py
@@ -28,6 +28,7 @@ from taskbroker_client.constants import (
     DEFAULT_REBALANCE_AFTER,
     DEFAULT_WORKER_HEALTH_CHECK_SEC_PER_TOUCH,
     DEFAULT_WORKER_QUEUE_SIZE,
+    DEFAULT_WORKER_WARMUP_TIMEOUT_SEC,
     MAX_BACKOFF_SECONDS_WHEN_HOST_UNAVAILABLE,
     WORKER_CHILD_JOIN_TIMEOUT_SEC,
 )
@@ -143,6 +144,8 @@ class PushTaskWorker:
         push_task_timeout: float = 5,
         update_in_batches: bool = False,
         skip_awaiting_futures: bool = True,
+        min_ready: int | None = None,
+        warmup_timeout: float = DEFAULT_WORKER_WARMUP_TIMEOUT_SEC,
     ) -> None:
         app = import_app(app_module)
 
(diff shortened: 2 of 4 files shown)


clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -364,6 +421,10 @@ def signal_handler(*args: Any) -> None:
             server.add_insecure_port(f"[::]:{self._grpc_port}")
             server.start()
 
+            # Hold NOT_SERVING until children are warm so the pod stays out of
+            # the NEG/readiness set while its child processes are still loading.
+            self._await_children_warm()
+
```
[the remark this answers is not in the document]
enochtangg: Addressed

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -310,6 +316,57 @@ def _stop_health_check_thread(self) -> None:
...
+        <= 0 disables the gate entirely.
+        """
+        min_ready = min(self._min_ready, self._concurrency)
+        if min_ready <= 0:
+            return
+
+        warmup_start = time.monotonic()
+        deadline = warmup_start + self._warmup_timeout
+        timed_out = False
+        while self.worker_pool.ready_count < min_ready:
```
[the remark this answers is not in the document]
enochtangg: I think this is acceptable as this doesn't happen frequently. Decrementing the ready_count could have a worse failure mode: a SIGKILLed or segfaulted child never runs its decrement, so a live gauge permanently undercounts and the pod could never reach min_ready, never serving until the 90s timeout.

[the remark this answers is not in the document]
george-sentry: Agreed. In most scenarios, there is no reason for a child process to exit on startup before receiving any tasks.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -143,6 +144,8 @@ def __init__(
         push_task_timeout: float = 5,
         update_in_batches: bool = False,
         skip_awaiting_futures: bool = True,
+        min_ready: int | None = None,
```
george-sentry: On one hand, I can see why having this lever could be useful. On the other hand, it adds an extra lever that needs to be configured properly. Why not simply wait for all the children to become ready?

clients/python/src/taskbroker_client/constants.py
```diff
@@ -57,6 +57,12 @@
 The number of gRPC requests before touching the health check file
 """
 
+DEFAULT_WORKER_WARMUP_TIMEOUT_SEC = 90.0
```
george-sentry, line 60: Note that we need to account for `livenessProbe` and `readinessProbe` values here. For example, if the `livenessProbe` is less than this timeout, then the pod may be restarted before the workers have warmed up since we don't set the status to SERVING until they are all ready.

Is this something we should worry about? If so, how should we deal with it?

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: You're right, we should reduce the number of levers if possible. The lever allows us to tolerate some level of stuck child threads and enables other threads to start working. All that is speculative for v1 though. I'll remove this as we can always add back if we find it useful in practice in prod.

clients/python/src/taskbroker_client/constants.py
enochtangg: Good call, k8s offers `startupProbe` so that liveness can't kill the pod mid-warmup. We would just need to make sure it comfortably exceeds `warmup_timeout`.

george-sentry (approved):

clients/python/src/taskbroker_client/constants.py
george-sentry: We can also use a different mechanism for the liveness probe (so that it succeeds right away) while keeping the existing health endpoint for the readiness probe (so that it's not considered ready until SERVING).

PR #732, opened 2026-06-29
feat(taskbroker): Better Child Management

george-sentry:
## Linear
Refs [STREAM-1269](https://linear.app/getsentry/issue/STREAM-1269/fix-throughput-dips-during-worker-child-process-recycling)

## Description
In #731, we deferred setting the status to `SERVING` for workers in push mode until all threads were warmed up because before, Kubernetes thought workers were ready to receive activations even though they weren't, causing significant throughput declines during worker redeployments.

Something similar happens when child processes **recycle**, or exit voluntarily because they have executed a certain, configurable number of tasks. This is meant to prevent memory leaks.

In some cases, child processes recycle at the same time, **decreasing throughput significantly** until all child processes are ready again. We can solve this problem by improving the child management process as follows.

* Spawn children and wait until they are all ready before advertising `SERVING`
* When a process has executed the maximum number of tasks specified, it does NOT exit right away - instead, it tells the parent that it's ready to exit and continues working
* The parent spawns a new child process to replace the old one
* Once the newly spawned child is ready, the one that wanted to exit is killed by the parent

This way, after startup, there are always _exactly_ `--concurrency` child processes at all times.

<p align="center">
image
</p>
<p align="center"><i>Before</i></p>

<p align="center">
image
</p>
<p align="center"><i>After</i></p>

### DIFF
--- a/clients/python/src/taskbroker_client/worker/worker.py
+++ b/clients/python/src/taskbroker_client/worker/worker.py
@@ -937,7 +937,7 @@ class TaskWorkerProcessingPool:
 
             while not self._shutdown_event.is_set():
                 # Read any events that may have come in since the last loop iteration
-                received: list[ChildMessage] = []
+                received: List[ChildMessage] = []
 
                 while True:
                     try:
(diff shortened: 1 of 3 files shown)


untitaker:
> When a process has executed the maximum number of tasks specified, it does NOT exit right away - instead, it tells the parent that it's ready to exit and continues working

I believe this may result in higher and less predictable memory usage. Isn't there a way to avoid all child processes restarting at the same time by adding some jitter to the restart interval?

george-sentry:
> I believe this may result in higher and less predictable memory usage. Isn't there a way to avoid all child processes restarting at the same time by adding some jitter to the restart interval?

You are right, that is a possible effect. And yes, it is possible to add jitter.

george-sentry:
Closed in favor of #735.

PR #733, opened 2026-06-29
ref(TaskProducer): free up the GIL more

bmckerry:
When enabling TaskProducer for sentry's `track_outcome()`, latency until the producer delivery callback was called spiked.
I'm assuming this is because:
- `track_outcome()` is higher throughput for producing than anything else I've enabled TaskProducer for
- the current implementation of TaskProducer/workerchild can hog the GIL

This PR attempts to fix this by:
- deleting the gauge metric that was recorded on every produce. This metric is still roughly covered by the workerchild's `taskworker.task.futures_produced` metric: https://github.com/getsentry/taskbroker/blob/main/clients/python/src/taskbroker_client/worker/workerchild.py#L532-L541
- adding a configurable sleep to each iteration of `check_task_future_completion` - previously this only paused if there were no producer futures from a task, this should free up the GIL more and allow us to find an optimal pause value per TaskWorker instance

### DIFF
--- a/clients/python/src/taskbroker_client/worker/workerchild.py
+++ b/clients/python/src/taskbroker_client/worker/workerchild.py
@@ -175,6 +175,7 @@ def child_process(
     processing_pool_name: str,
     process_type: str,
     skip_awaiting_futures: bool,
+    future_checking_frequency: float,
     ready_counter: "Synchronized[int] | None" = None,
 ) -> None:
     """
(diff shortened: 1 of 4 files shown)


clients/python/src/taskbroker_client/worker/workerchild.py
untitaker, line 320: I wonder if you want to use `concurrent.futures.wait` here

untitaker (approved):

clients/python/src/taskbroker_client/worker/workerchild.py
bmckerry: I'll take a look at other improvements such as this in a later PR

PR #737, opened 2026-06-30
feat(killswitch): argument-based fine-grained killswitches

untitaker:
ref STREAM-1276

## Why
Today we can only killswitch a task by name — all-or-nothing. To drop traffic for a single project mid-incident we either nuke the whole task or wait on app devs to add a `sentry.killswitches` definition (came out of INC-2274).

## What
`drop_task_killswitch` entries can now match on a task's arguments:

```yaml
drop_task_killswitch:
  - sentry.tasks.unmerge            # drop the whole task (as before)
  - name: sentry.tasks.unmerge      # or drop only when args match
    select_arg:   {0: 123}          # 1st positional arg == 123, OR
    select_kwarg: {project_id: 123} # kwarg project_id == 123
```

A rule drops the task if **any** condition matches. Enforced at ingest and in the upkeep DB-cleanup path. Matching reads the msgpack `parameters_bytes` (zstd-aware) and **fails open** on any decode error, so a changed task signature can't drop unrelated traffic. [Run book](https://www.notion.so/1b48b10e4b5d80a3911bc0b019ba73b5) updated.

### DIFF
--- a/src/killswitch.rs
+++ b/src/killswitch.rs
@@ -1,0 +1,351 @@
+//! Fine-grained task killswitches.
+//!
+//! The `drop_task_killswitch` runtime config is a list of rules. A rule is either:
+//!
+//! * a bare task name (`KillswitchRule::Name`) — drop every activation of that task, or
+//! * a selector (`KillswitchRule::Selector`) — drop an activation only when its arguments
+//!   match, so we can e.g. drop `sentry.tasks.unmerge` for a single `project_id`.
+//!
+//! Selectors match against the activation's `parameters_bytes`, which is msgpack-encoded
+//! as `{"args": [...], "kwargs": {...}}` and may be zstd-compressed (see
+//! `headers["compression-type"]`). All parsing **fails open**: any decode error means the
+//! task is *not* dropped, so a killswitch can never take down traffic it wasn't meant to.
+
+use std::collections::HashMap;
+
+use sentry_protos::taskbroker::v1::TaskActivation;
+use serde::Deserialize;
+
+use crate::kafka::deserialize_raw::{COMPRESSION_HEADER, COMPRESSION_ZSTD};
+
+/// Upper bound for decompressing `parameters_bytes`, mirroring the worker's limit in
+/// `deserialize_raw`.
+const MAX_DECOMPRESSED_SIZE: usize = 64 * 1024 * 1024;
+
+/// A single entry in `drop_task_killswitch`.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+#[serde(untagged)]
+pub enum KillswitchRule {
+    /// A bare task name: drop every activation of this task.
+    Name(String),
+    /// An argument selector: drop only activations whose arguments match.
+    Selector(KillswitchSelector),
+}
+
+/// Drop activations of `name` whose arguments match any of the listed conditions.
+#[derive(Debug, Deserialize, Clone, PartialEq)]
+pub struct KillswitchSelector {
+    /// The task name this selector applies to.
+    pub name: String,
+    /// Match by positional argument index, e.g. `{0: 123}` for "first arg == 123".
+    #[serde(default)]
+    pub select_arg: HashMap<usize, serde_yaml::Value>,
+    /// Match by keyword argument name, e.g. `{project_id: 123}`.
+    #[serde(default)]
+    pub select_kwarg: HashMap<String, serde_yaml::Value>,
+}
+
+impl KillswitchRule {
+    /// The task name this rule applies to, regardless of variant.
+    pub fn taskname(&self) -> &str {
+        match self {
+            KillswitchRule::Name(name) => name,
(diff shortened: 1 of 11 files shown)


untitaker:
I'm not super happy with how heavy upkeep is getting here, but i don't think we want to put this into push workers or the grpc path either. I think maybe we want to have a way to disable upkeep killswitches, wdyt?

george-sentry (approved):

src/killswitch.rs
```diff
@@ -0,0 +1,351 @@
...
+    {
+        return Some("whole_task");
+    }
+
+    // Selector killswitch: only decode the activation when a selector targets this task.
+    let has_selector = rules
+        .iter()
+        .any(|rule| matches!(rule, KillswitchRule::Selector(sel) if sel.name == taskname));
+    if has_selector
+        && let Ok(activation) = TaskActivation::decode(activation_bytes)
```
evanh, line 152: I'm a little worried about the overhead from this. Do we think we'd have a problem if someone decided to killswitch a specific project in the span/transaction tasks?

untitaker: i think upkeep will run slower for sure, not sure though that throughput decreases.

i think during an inc, the pool is generally already in a bad state though.

we could try the killswitch out for a non-existent project (in s4s2) to see how it will affect throughput

enochtangg (commented):
> I'm not super happy with how heavy upkeep is getting here, but i don't think we want to put this into push workers or the grpc path either. I think maybe we want to have a way to disable upkeep killswitches, wdyt?

I agree it's difficult to tell what the throughput implications are without some load testing. If it turns out to be a bigger issue, I wonder if it's feasible to drop the deletion scan from upkeep and accept that taskbroker will need to still run the remaining tasks in store. Alternatively, what if we somehow only ran the deletion once by tracking the last seen of the runtime config?

src/killswitch.rs
markstory: This will add overhead while a killswitch is active. But at the same time we only use killswitches when something bad/worse is also happening.

untitaker:
@markstory 

> This will add overhead while a killswitch is active. But at the same time we only use killswitches when something bad/worse is also happening.

that's my thinking here with perf as well.

how does this sound?

1. merge this PR -- merging and deploying alone doesn't have impact, since the killswitch is not enabled
2. we turn on a killswitch in s4s2 and observe
3. if we see issues, reconvene

but i think it'll be fine, if we are killswitching the pool is already in a bad state.

@enochtangg 

> I wonder if it's feasible to drop the deletion scan from upkeep and accept that taskbroker will need to still run the remain tasks in store.

i think we can do that, would certainly simplify the code a lot. but then we ought to do it for both killswitches for consistency...

> Alternatively, what if we somehow only ran the deletion once by tracking the last seen of the runtime config?

i think the code is already too complicated tbh

src/killswitch.rs
evanh: OK. It feels like there are scenarios where the killswitches could make things worse, like if broker CPU is already high. Something to explore when testing.

markstory:
> i think we can do that, would certainly simplify the code a lot. but then we ought to do it for both killswitches for consistency..

While simpler, this will extend the time it takes to recover from a task that has been killswitched as we'll need to process/timeout all the activations that made it into the database already.

evanh:
> how does this sound?
> 
>     1. merge this PR -- merging and deploying alone doesn't have impact, since the killswitch is not enabled
> 
>     2. we turn on a killswitch in s4s2 and observe
> 
>     3. if we see issues, reconvene
> 
> 
> but i think it'll be fine, if we are killswitching the pool is already in a bad state.

I think this generally makes sense. My worry is that if the pool is in a specific bad state, killswitching might actually make things worse (e.g. brokers are CPU loaded, then we add killswitches). But I don't think killswitches would be used for that kind of problem anyways.

Could you add some metrics around the timing and killswitching (if they don't already exist). E.g. upkeep time spent killswitching, maybe a duration metric for how long the broker spends processing killswitches?

untitaker:
added a missing timing for the ingest path (the timing itself is also not emitted if there are no killswitches active, for the sake of performance)

src/upkeep.rs
```diff
@@ -302,13 +302,26 @@ pub async fn do_upkeep(
...
+            result_context.killswitched += count;
+        }
+        // Selector rules require decoding each candidate activation's arguments, so they
+        // use a separate scan-and-delete path.
+        if !killswitched_selectors.is_empty()
+            && let Ok(count) = store
+                .remove_killswitched_selectors(killswitched_selectors)
+                .await
+        {
+            result_context.killswitched += count;
```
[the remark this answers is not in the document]
untitaker: i thikn that's not a concern. we will block updates of runtime config during this period but not other high-throughput processes, but this is maybe a flaw in runtime config which we should fix

evanh (approved):

PR #739, opened 2026-07-01
Silence error logging when exception is silenced

cmanallen:
Silenced exceptions are emitting `NoRetriesRemainingError` errors in Sentry. If the error is silenced, the fact that we've exhausted the retry quota is expected behavior.

Fixes [SENTRY-5PM9](https://sentry.sentry.io/issues/7473353475/?project=1&query=is%3Aunresolved&referrer=issue-stream&sort=freq)


untitaker (commented):
so does that mean you have a silenced exception that is also configured to trigger retries?

cmanallen:
@untitaker The retry list is a subset of the silenced list: https://github.com/getsentry/sentry/blob/168eda1ff479d8131237c3198b385058c87a67cf/src/sentry/sentry_apps/tasks/sentry_apps.py#L178-L190.

untitaker (approved):

PR #741, opened 2026-07-03
feat(taskbroker-client): expose task enqueue producer future

lvthanh03:
Refs STREAM-1309

Modifies `TaskNamespace.send_task()` to return the existing producer future, and adds `Task.apply_async_with_future()` for dispatching a task and also returning that producer future.

This lets callers coordinate task enqueue delivery; in the case of the spans buffer enqueueing the process_segment task, it needs to know that the produce succeeded before calling `done_flush_segments`.

### DIFF
--- a/clients/python/src/taskbroker_client/task.py
+++ b/clients/python/src/taskbroker_client/task.py
@@ -27,7 +27,7 @@ from taskbroker_client.constants import (
 from taskbroker_client.retry import Retry
 
 if TYPE_CHECKING:
-    from taskbroker_client.registry import TaskNamespace
+    from taskbroker_client.registry import ProducerFuture, TaskNamespace
 
 
 ALWAYS_EAGER = False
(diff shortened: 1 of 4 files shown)


clients/python/src/taskbroker_client/task.py
```diff
@@ -185,6 +198,31 @@ def apply_async(
                 wait_for_delivery=self.wait_for_delivery,
             )
 
+    def apply_async_with_future(
```
untitaker, line 201: why can't we make apply_async return the future? It's not a breaking change to do so

untitaker (approved):

clients/python/src/taskbroker_client/task.py
lvthanh03: I thought there might be some risk in changing the public client API contract for _every_ task dispatch, though I just checked in sentry and there isn't any code that uses the return value of `apply_async` so we won't break any exisitng logic if we make apply async return the future.

Either way, I think it's clearer to only return the producer future when the user explicitly asks for it back.

bmckerry (approved):

evanh (approved):

PR #746, opened 2026-07-06
test(integration): Add postgres-backed push mode integration tests

evanh:
## Summary

Adds integration test coverage for **postgres-backed push mode**. Every existing integration test runs pull mode against SQLite, so the push pipeline (the broker claims activations from the store and pushes them to worker gRPC endpoints) had no end-to-end coverage. This suite launches a real `taskbroker` binary in push mode backed by the devservices postgres.

Most of the existing Python integration-test infrastructure is reused (broker subprocess lifecycle, config-as-YAML, kafka helpers, `get_available_ports`, Makefile/CI per-test job structure). Postgres was already a devservices dependency, and the broker self-provisions the DB + runs migrations on startup, so no external DB setup is needed.

## What's new

- **`push_worker.py`** — a minimal `MockWorkerServicer` gRPC server implementing `WorkerService.PushTask`. This is the key new piece: in push mode the broker is the gRPC *client*, whereas the existing `worker.py` is a pull-mode client. The mock records pushed activation IDs and can be told to `fail` (abort) or `hang`.
- **`helpers.py`** — `PostgresConfig` dataclass (with `.store_config()`) plus `get_num_tasks_in_postgres[_by_status/_group_by_status]`, `drop_postgres_db`, and `unique_pg_database_name`, mirroring the existing sqlite helpers. Uses `psycopg2-binary`.
- **`test_push_postgres.py`** — three tests:
  - **happy-path**: healthy worker → every task pushed exactly once and marked `Processing` in postgres.
  - **push-failure revert**: rejecting worker → nothing marked processing, no tasks lost, claimed tasks revert to pending (push-thread revert path).
  - **claim-expiration / upkeep**: hanging worker saturates the push queue → claimed tasks are stranded and upkeep reverts them to pending (verified via the `claim_expiration_reset` upkeep log signal).
- **`Makefile` + CI** — a `test-push-postgres` target and a matching `push-postgres-integration-test` job (copied from the multi-topic job).

## Notes

- Uses `psycopg2-binary` rather than `psycopg` (v3) because the devinfra package index doesn't have v3.
- A few things surfaced while writing these that are worth knowing: the new config format requires the deadletter topic to be declared in `kafka_topics`, and the postgres store persists **capitalized** status enum names (`Pending`/`Claimed`/`Processing`) unlike sqlite's lowercase.
- Following the `test-multi-topic` precedent, the new target is a standalone CI job and is **not** added to the `integration-test` aggregate.

## Test plan

- `make test-push-postgres` (or `uv run --active python -m pytest integration_tests/integration_tests/test_push_postgres.py`) — all 3 tests pass locally (~105s) against devservices kafka + postgres.
- `black`, `isort`, `flake8`, and `mypy` are clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -528,3 +528,41 @@ jobs:
       - name: Run multi-topic integration test
         run: |
           make test-multi-topic
+
+  push-postgres-integration-test:
+    name: Push mode postgres integration test
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
+
+      - name: Install cmake
+        uses: lukka/get-cmake@28983e0d3955dba2bb0a6810caae0c6cf268ec0c # latest
+
+      - name: Install libcurl
+        run: sudo apt-get update && sudo apt-get install -y libcurl4-openssl-dev
+
+      - uses: actions-rs/toolchain@16499b5e05bf2e26879000db0c1d13f7e13fa3af # pin@v1
+        with:
+          toolchain: stable
+          profile: minimal
+          override: true
+
+      - uses: swatinem/rust-cache@81d053bdb0871dcd3f10763c8cc60d0adc41762b # pin@v1
+        with:
+          key: ${{ github.job }}
+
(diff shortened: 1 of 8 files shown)


enochtangg (approved):
Nice stuff!

PR #756, opened 2026-07-16
fix(taskbroker): Replace `worker_map` Default Instead of Merging

george-sentry:
For some reason, the Seer push taskbroker was getting a `connection refused` error when trying to reach the taskworker service. After some deep investigation, I found that Figment is very silly. Instead of replacing `worker_map` with whatever the user provides, it _merges_ it.

So if you provide `TASKBROKER_WORKER_MAP={seer=foo}`, its value will be `{sentry=http://127.0.0.1:50052, seer=foo}` rather than just `{seer=foo}`. I strongly suspect that is the cause.

I have changed the code to replace `worker_map` with whatever the user provides, falling back on the previous default if the user doesn't provide anything.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -318,14 +318,19 @@ async fn main() -> Result<(), Error> {
             let mut map = HashMap::new();
 
             for (application, endpoint) in config.worker_map.clone() {
-                let worker = match Worker::connect(config.clone(), endpoint).await {
+                let worker = match Worker::connect(config.clone(), endpoint.clone()).await {
                     Ok(w) => {
-                        debug!("Connected to worker!");
+                        debug!(application, endpoint, "Connected to worker!");
                         Box::new(w) as Box<dyn WorkerClient>
                     }
 
                     Err(e) => {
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -318,14 +318,19 @@ async fn main() -> Result<(), Error> {
             let mut map = HashMap::new();
 
             for (application, endpoint) in config.worker_map.clone() {
-                let worker = match Worker::connect(config.clone(), endpoint).await {
+                let worker = match Worker::connect(config.clone(), endpoint.clone()).await {
```
evanh, line 321: This is a good point, can we have the startup fail if the worker map is empty?

evanh (approved):

enochtangg (approved):

src/main.rs
george-sentry: Yepp, I added a `validate` attribute to the `worker_map` field to make sure it has at least one entry. Also, I changed the code to use a default when the user doesn't provide anything. But this time, the user's `worker_map` replaces the default instead of merging with it.

PR #764, opened 2026-07-23
fix(python): Shut workers down without raising from signal handlers

untitaker:
Replaces #762.

## The bug

Both worker signal handlers raised `KeyboardInterrupt`. That unwinds at an arbitrary bytecode, and the worker's shutdown event is a `multiprocessing.Event` that both workers also use as their backoff sleep (`fetch_task`, `_send_update_task`, `_send_results`).

Interrupting `multiprocessing.Event.wait()` lands inside `Condition.wait()`, which releases the lock N times, sleeps on a semaphore, then reacquires in a `finally`. An interrupt during that `finally` leaves the lock unheld while `Event.wait`'s enclosing `with` block still releases it, or leaves `_sleeping_count`/`_woken_count` skewed so the next `notify()` trips one of its three asserts.

## Why not just set the event from the handler

That is what #762 does, and it swaps the corruption for a hang. `Event.set()` — both the threading and multiprocessing flavours — takes a non-reentrant lock. A signal arriving while the same thread is inside `wait()` holding that lock makes the handler block on a lock only that thread can release. The process is then unkillable by SIGTERM, so k8s waits out the grace period and SIGKILLs, losing exactly the in-flight tasks the change was meant to protect. The multiprocessing flavour is worse still: another process can be holding it.

## The fix

Python runs signal handlers on the main thread between bytecodes, so a plain attribute assignment is safe and anything that takes a lock is not. Handlers now only do `self._shutdown_signal.request()`, which assigns a bool. The loops act on it.

- The worker shutdown event becomes `ShutdownSignal`, wrapping a `threading.Event`. It was never shared with child processes — unlike the pool's own `_shutdown_event` — so it never needed to be a `multiprocessing.Event`.
- Sleeps poll the flag every `SHUTDOWN_POLL_INTERVAL_SEC` (0.5s). The event is still there so a shutdown noticed on the main thread wakes the result thread immediately, and is only ever set from normal code.
- The push worker polls `server.wait_for_termination(timeout=...)` instead of relying on a `server.stop()` call from inside the handler. grpc's `stop()` takes its own locks and is not safe there either.
- The pull worker returns 0 instead of re-raising `KeyboardInterrupt`, so `cli.py` gets an exit code.
- The push worker skips `server.start()` if shutdown was already requested, and only stops a server it started.
- A pull worker that is shutting down no longer asks for another task, either in `fetch_task()` or via `update_task`'s `fetch_next`.

The last three are carried over from #762.

## Known gap

A SIGTERM arriving during a blocking `get_task()` or the client's all-hosts-unavailable `time.sleep()` (`worker/client.py:371`, up to 20s) is not noticed until that call returns. The old `KeyboardInterrupt` did cut those short. Fixing it properly means putting a deadline on those calls, which is a separate change.

## Tests

`ShutdownSignal` unit tests cover the timeout, both wakeup paths, and a real SIGALRM handler firing while `wait()` is sleeping. Worker tests cover clean SIGTERM exit in pull and push mode, shutdown before `server.start()`, shutdown during warmup, and the two fetch-suppression paths.

ref SENTRY-5MB5

### DIFF
--- a/clients/python/tests/worker/test_worker.py
+++ b/clients/python/tests/worker/test_worker.py
@@ -30,6 +30,7 @@ from sentry_protos.taskbroker.v1.taskbroker_pb2 import (
     TASK_ACTIVATION_STATUS_COMPLETE,
     TASK_ACTIVATION_STATUS_FAILURE,
     TASK_ACTIVATION_STATUS_RETRY,
+    FetchNextTask,
     PushTaskRequest,
     PushTaskResponse,
     RetryState,
@@ -45,6 +46,7 @@ from taskbroker_client.types import InflightTaskActivation, ProcessingResult
 from taskbroker_client.worker.producer import TaskProducer, _pending_futures
 from taskbroker_client.worker.worker import (
     PushTaskWorker,
+    ShutdownSignal,
     TaskWorker,
     TaskWorkerProcessingPool,
     WorkerServicer,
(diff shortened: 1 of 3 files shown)


clients/python/tests/worker/test_worker.py
```diff
@@ -831,29 +961,126 @@ def warm_up() -> None:
     assert timeout_calls == []
 
 
-def test_start_does_not_serve_when_shutdown_during_warmup() -> None:
-    from grpc_health.v1 import health_pb2
+@contextlib.contextmanager
```
evanh, line 964: This is some complicated testing. I'm OK with it but we are getting to edge of "human-readable".

evanh (approved):

PR #767, opened 2026-07-27
ref(o11y): Support `sentry-sdk` configured with the streaming trace lifecycle

alexander-alderman-webb:
Use the `sentry_sdk.traces` API when the streaming lifecycle is enabled.
The previous API is a no-op when the lifecycle is enabled.

This PR consists of:
- cherry-picking https://github.com/getsentry/taskbroker/commit/3db8d2d6cc9e4b84af4a53d3b3c74710be4e7b89
- an early return of `nullcontext()` when `start_span()` is called without an active span to become the parent. See https://github.com/getsentry/taskbroker/commit/d6885e4b35b7ef27d850397ee46ce48445274b81

We saw a large increase segment spans in the previous dogfooding attempt. The majority of the increased span volume consisted of spans with op `queue.publish` and the name `sentry.spans.process_segments.process_segment`. These new spans are segment spans and were previously dropped as the existing `start_span()` API is a no-op when there is no transaction.

The early return is added to keep span volume constant when switching the trace lifecycle while using `taskbroker`.


clients/python/src/taskbroker_client/worker/workerchild.py
```diff
@@ -612,45 +614,49 @@ def _execute_activation(
...
+                    "taskworker": {
+                        "task": activation.taskname,
+                    }
+                },
+            ) as transaction,
         ):
-            transaction.set_data(
-                "taskworker-task", {"args": args, "kwargs": kwargs, "id": activation.id}
-            )
+            # Do not attach on StreamedSpan because eager serialization increases memory use.
```
ericapisani, line 639: More because I'm curious - how significant is the increase in memory due to the eager serialization?

clients/python/src/taskbroker_client/sdk.py
```diff
@@ -0,0 +1,88 @@
...
+
+        span = sentry_sdk.start_transaction(transaction, custom_sampling_context=sampling_context)
+        for key, value in attributes.items():
+            span.set_data(key, value)
+    except Exception:
+        pass
+
+    if span is None:
+        return nullcontext()
+    return span
```
ericapisani, line 51: Looking at this method and the `start_span` method below - I don't fully understand what types of exceptions we're guarding against (and from where, the streamed span or transaction logic) that we need this null context. 

Could you provide some more context on why we need the try/except above along with this fallback?

clients/python/src/taskbroker_client/sdk.py
```diff
@@ -0,0 +1,88 @@
...
+
+        if is_span_streaming:
+            return sentry_sdk.traces.start_span(
+                name=name,
+                attributes={
+                    "sentry.op": op,
+                    "sentry.origin": origin,
+                    **attributes,
+                },
+            )
```
ericapisani, line 74: This can also be written as:


```suggestion
        if is_span_streaming:
            if sentry_sdk.traces.get_current_span() is None:
                        return nullcontext()
            
            return sentry_sdk.traces.start_span(
                name=name,
                attributes={
                    "sentry.op": op,
                    "sentry.origin": origin,
                    **attributes,
                },
            )
```

clients/python/src/taskbroker_client/sdk.py
alexander-alderman-webb: We used a shim without the `try...except` in `sentry` itself. There's nothing in particular that I'd expect to blow up. We could also get rid of the `try...except` 🤷

clients/python/src/taskbroker_client/worker/workerchild.py
alexander-alderman-webb: Context is here: https://github.com/getsentry/taskbroker/pull/757#issuecomment-5059161880

clients/python/src/taskbroker_client/sdk.py
alexander-alderman-webb: The suggestion would increase indentation, which is undesirable in Python.

evanh (approved):

PR #773, opened 2026-08-04
feat(taskbroker): Add Tokio worker-thread utilization metric

enochtangg:
## Summary

Refs: STREAM-1672.

Adds per-broker Tokio runtime metrics so operators can see how much of the worker-thread pool is actually busy, rather than inferring it from CPU.

Context: in the s4s2 process-segments-push investigation the broker was starved on Tokio worker threads while CPU sat < 1 core, because the push workload is I/O-bound. We currently have no way to see this. This exposes utilization (how much of the pool is in use), not just saturation.

Prior art: Relay emits the same family of Tokio runtime metrics via `tokio::runtime::RuntimeMetrics`.

## Changes
Adds a lightweight sampler that reads `Handle::current().metrics()` on an interval and emits:

| Metric | Type | Meaning |
|---|---|---|
| `runtime.worker_utilization` | gauge (0–1) | mean worker-thread busy ratio (the primary signal) |
| `runtime.global_queue_depth` | gauge | tasks in the runtime's global (injection) queue |
| `runtime.worker_local_queue_depth` | gauge | tasks summed across workers' local run queues |
| `runtime.num_workers` | gauge | worker-thread count |
| `runtime.num_alive_tasks` | gauge | live tasks in the runtime |

This change requires the `tokio_unstable` config. The per-worker fields are only exposed under `--cfg tokio_unstable`, so this PR enables it. It's a compile-time flag which shouldn't have runtime overhead as Tokio collects these counters regardless. It is also the same approach Relay runs
in production.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,9 @@
+[build]
+# Exposes Tokio's per-worker runtime metrics (worker_total_busy_duration,
+# worker_local_queue_depth, etc.) via `Handle::metrics()`. These are gated
+# behind the `tokio_unstable` cfg.
+rustflags = ["--cfg", "tokio_unstable"]
+
 [env]
 # Workaround for https://github.com/confluentinc/librdkafka/pull/5012
 CMAKE_POLICY_VERSION_MINIMUM = "3.10"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -3,6 +3,9 @@
 # worker_local_queue_depth, etc.) via `Handle::metrics()`. These are gated
 # behind the `tokio_unstable` cfg.
 rustflags = ["--cfg", "tokio_unstable"]
+# Doc-tests are compiled by rustdoc, which reads rustdocflags rather than
+# rustflags; without this they'd miss the cfg above and hit our compile_error!.
+rustdocflags = ["--cfg", "tokio_unstable"]
 
 [env]
 # Workaround for https://github.com/confluentinc/librdkafka/pull/5012
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -150,6 +150,11 @@ pub struct Config {
     /// (reclaiming free pages) are executed
     pub maintenance_task_interval_ms: u64,
 
+    /// The frequency (in milliseconds) at which Tokio runtime metrics
+    /// (worker-thread utilization, queue depths) are sampled and emitted.
+    #[validate(range(min = 1))]
+    pub runtime_metrics_interval_ms: u64,
+
     /// The maximum number of seconds a task can be delayed until.
     /// Tasks delayed greater than this duration are capped.
     pub max_delayed_task_allowed_sec: u64,
@@ -268,6 +273,7 @@ impl Default for Config {
             health_check_killswitched: false,
             upkeep_deadline_reset_skip_after_startup_sec: 60,
             maintenance_task_interval_ms: 6000,
+            runtime_metrics_interval_ms: 5000,
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 5000000,
             grpc_max_message_size: 52 * 1024 * 1024, // 52MB
--- a/src/config/mod.rs
+++ b/src/config/mod.rs
@@ -150,10 +150,12 @@ pub struct Config {
     /// (reclaiming free pages) are executed
     pub maintenance_task_interval_ms: u64,
 
-    /// The frequency (in milliseconds) at which Tokio runtime metrics
-    /// (worker-thread utilization, queue depths) are sampled and emitted.
-    #[validate(range(min = 1))]
-    pub runtime_metrics_interval_ms: u64,
+    /// The frequency at which Tokio runtime metrics (worker-thread
+    /// utilization, queue depths) are sampled and emitted, e.g. `"1s"` or
+    /// `"500ms"`.
+    #[serde(with = "crate::serde::duration")]
+    #[validate(custom(function = "validate::nonzero_duration"))]
+    pub runtime_metrics_interval: Duration,
 
     /// The maximum number of seconds a task can be delayed until.
     /// Tasks delayed greater than this duration are capped.
@@ -273,7 +275,7 @@ impl Default for Config {
             health_check_killswitched: false,
             upkeep_deadline_reset_skip_after_startup_sec: 60,
             maintenance_task_interval_ms: 6000,
-            runtime_metrics_interval_ms: 5000,
+            runtime_metrics_interval: Duration::from_secs(1),
             max_delayed_task_allowed_sec: 3600,
             max_message_size: 5000000,
             grpc_max_message_size: 52 * 1024 * 1024, // 52MB
--- a/src/runtime_metrics.rs
+++ b/src/runtime_metrics.rs
@@ -1,20 +1,38 @@
-use std::time::Duration;
+//! Per-broker Tokio runtime metrics.
+//!
+//! Emits gauges describing how busy the shared Tokio worker pool is, the most
+//! important being `runtime.worker_utilization`: the mean fraction of worker
+//! threads that were actively executing tasks over the last sampling window.
+//!
+//! Tokio exposes `worker_total_busy_duration` as a **monotonic cumulative counter** per worker,
+//! not an instantaneous rate. A single reading is meaningless on its own; to derive
+//! a utilization ratio we remember the previous cumulative total (and when it was taken) and divide
+//! the delta over the window by that window's maximum possible busy time
+//! (`elapsed * num_workers`). [`WorkerBusySampler`] owns that state.
 
-use tokio::runtime::Handle;
+use std::time::{Duration, Instant};
+
+use tokio::runtime::{Handle, RuntimeMetrics};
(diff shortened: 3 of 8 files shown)


.cargo/config.toml
```diff
@@ -1,3 +1,9 @@
+[build]
+# Exposes Tokio's per-worker runtime metrics (worker_total_busy_duration,
+# worker_local_queue_depth, etc.) via `Handle::metrics()`. These are gated
+# behind the `tokio_unstable` cfg.
+rustflags = ["--cfg", "tokio_unstable"]
```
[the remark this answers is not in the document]
enochtangg: Copied `.cargo` into the build stage before the dependency build so tokio is compiled with the cfg

fpacifici:
> tasks summed across workers' local run queues

What does this represent ?

src/config/mod.rs
```diff
@@ -150,6 +150,11 @@ pub struct Config {
     /// (reclaiming free pages) are executed
     pub maintenance_task_interval_ms: u64,
 
+    /// The frequency (in milliseconds) at which Tokio runtime metrics
+    /// (worker-thread utilization, queue depths) are sampled and emitted.
+    #[validate(range(min = 1))]
+    pub runtime_metrics_interval_ms: u64,
```
george-sentry: I created custom serializers / deserializers for durations so that users can write something like "500ms" or "2s". For example, see the `backoff` field in `FetchConfig`.

```rs
pub struct FetchConfig {
    ...
    #[serde(with = "crate::serde::duration")]
    pub backoff: Duration,
    ...
}
```

The advantage is that there's no need to convert into a duration as it's already one. On the other hand, by making it a number and using the `_ms` suffix, it will always be obvious that it must be milliseconds.

Just throwing this out, feel free to use whichever way seems best 😁

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+    let runtime_metrics = Handle::current().metrics();
+
+    let mut timer = time::interval(interval);
+    timer.set_missed_tick_behavior(MissedTickBehavior::Skip);
+
+    #[cfg(not(tokio_unstable))]
+    tracing::warn!(
+        "Tokio worker utilization metric unavailable: built without `--cfg tokio_unstable`; \
+         emitting stable runtime counts only"
+    );
```
george-sentry: Do you think it would be reasonable to simply require that `tokio_unstable` always be enabled?

src/config/mod.rs
```diff
@@ -268,6 +273,7 @@ impl Default for Config {
             health_check_killswitched: false,
             upkeep_deadline_reset_skip_after_startup_sec: 60,
             maintenance_task_interval_ms: 6000,
+            runtime_metrics_interval_ms: 5000,
```
fpacifici: I think you can reduce this to 1 second. 
Recording metrics every second means every several thousands of tasks. it is ok. 
When you go above 1 second you start seeing spiky metrics as they become sparse.

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+
+    use super::utilization_ratio;
+
+    /// A single sampling result derived from the runtime's per-worker metrics.
+    pub struct Sample {
+        /// Mean worker-thread busy ratio in `[0.0, 1.0]`.
+        pub utilization: f64,
+        /// Number of tasks queued in the runtime's global (injection) queue.
+        pub global_queue_depth: usize,
+        /// Total tasks queued across all workers' local run queues.
```
fpacifici: What does this mean ? Is it the sum of all the queue length for each task ?

src/runtime_metrics.rs
fpacifici, line 1: I think you need some docs for this module that explains the approach to exrtact the busy time. It is hard to infer the rationale for computing the metric this way over simply probing the gauge periodically.

I think the key is that you compute the busy time in intervals because Tokio only gives you the busy time from the beginning of time and you need to log the metric periodically, thus you need this concept of sampling and interval.

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+        pub local_queue_depth: usize,
+    }
+
+    /// Tracks the cumulative worker busy duration between samples so that
+    /// utilization can be computed from the delta over the sampling window.
```
fpacifici: What's the "delta over the sampling window" ?

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+    }
+
+    /// Sums `worker_total_busy_duration` across every worker thread, in seconds.
+    fn total_busy_secs(runtime_metrics: &RuntimeMetrics) -> f64 {
+        (0..runtime_metrics.num_workers())
+            .map(|worker| {
+                runtime_metrics
+                    .worker_total_busy_duration(worker)
```
fpacifici: Does this return the busy duration from when the task is created ?

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+
+            Sample {
+                utilization: utilization_ratio(busy_delta_secs, elapsed_secs, num_workers),
+                global_queue_depth: runtime_metrics.global_queue_depth(),
+                local_queue_depth,
+            }
+        }
+    }
+
+    /// Sums `worker_total_busy_duration` across every worker thread, in seconds.
```
fpacifici: This is really easy to infer from the code. If you want to add a comment, it would be more useful to explain what the meaning of this number is for your module: "Computes the total time spent busy by all tasks from the creation till now"

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+            }
+        }
+
+        pub fn sample(&mut self, runtime_metrics: &RuntimeMetrics) -> Sample {
+            let now = Instant::now();
+            let busy_secs = total_busy_secs(runtime_metrics);
+            let elapsed_secs = now.duration_since(self.last_sample).as_secs_f64();
+            // `worker_total_busy_duration` is monotonic, but guard against clock
+            // effects / worker count changes producing a negative delta.
+            let busy_delta_secs = (busy_secs - self.prev_busy_secs).max(0.0);
```
fpacifici: How can this be lower than 0 ?

src/runtime_metrics.rs
```diff
@@ -0,0 +1,170 @@
...
+use tokio::time::{self, MissedTickBehavior};
+
+/// Computes the mean worker-thread busy ratio in `[0.0, 1.0]`.
+///
+/// `busy_delta_secs` is the total busy time accrued across all worker threads
+/// since the previous sample; dividing by `elapsed_secs * num_workers` (the
+/// maximum possible busy time for the whole pool over that window) yields the
+/// fraction of the pool that was busy.
+#[cfg(tokio_unstable)]
+fn utilization_ratio(busy_delta_secs: f64, elapsed_secs: f64, num_workers: usize) -> f64 {
```
fpacifici: nit: rename `busy_delta_secs` into something like `busy_secs`. 
While it is reasonable to specify this is a delta, this method does not care about the detla.
Reading busy_delta_secs I think this is a delta in seconds between two events, while it is the busy time.

fpacifici (approved):
Please address the comments before merging. There is quite some improvement that can be made on the clarity of this system

src/config/mod.rs
enochtangg: Thanks for adding that, updated to use duration directly.

src/runtime_metrics.rs
enochtangg: I think so, updated to explicitly require unstable

src/runtime_metrics.rs
enochtangg: Not tasks, the threads. Tokio's scheduler gives each worker (OS) thread its own local run queue; runnable tasks are scheduled onto a worker's local queue. This metric is the sum of how many are currently sitting in each worker's queue. This essentially tells us how many tasks are backed up waiting to be run

src/runtime_metrics.rs
enochtangg: Good idea, updated docs

src/runtime_metrics.rs
enochtangg: In this case, the sampling window is the time between two consecutive samples. Since Tokio only exposes `worker_total_busy_duration` as a counter that grows from runtime start, so there's no instantaneous "busy now" reading. Each tick, we're store the previous cumulative total and subtract it from the current one; that difference is the busy time accrued during this window, which I then divide by the window's capacity (num_workers × elapsed) to get utilization. I'll make this explicit in the module docs.

src/runtime_metrics.rs
enochtangg: It's the cumulative time that worker thread has spent busy polling tasks since the runtime started

src/runtime_metrics.rs
enochtangg: Will do, updated

src/runtime_metrics.rs
enochtangg: In practice it can't. The `worker_total_busy_duration` is monotonic and the worker count is fixed after startup, so the current cumulative total is always ≥ the previous. The .max(0.0) is just defensive.

PR #774, opened 2026-08-04
fix(python): Make worker occupancy time-weighted and warm-aware

enochtangg:
Refs: STREAM-1680

The `taskworker.worker.occupancy` under-reported real utilization, especially during rollouts, it stayed low even when workers were saturated and a large backlog was draining.

Changes:
1. Time-weighted busy threads: Started tracking busy time  instead of a boolean.  Each flush sums busy-seconds per child, closing any open segment at the interval boundary and reopening it, so tasks longer than the interval keep contributing. Occupancy is now `busy_time / (elapsed * running_children)`, the time-weighted average of concurrently-busy children over available capacity.
2. Warm-gated emission: Occupancy is only emitted once at least one child is `running`, so warming pods no longer publish zeros when warming.
3. running denominator: Divides by warmed-up children rather than target concurrency.

Testing
- Unit tests, see `test_worker.py`.
- Verified locally end-to-end: broker + push worker with `--prometheus-port`. Test produced `examples` task load and scraped `taskworker_worker_occupancy`. Occupancy tracks expected fractions.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/worker.py
+++ b/clients/python/src/taskbroker_client/worker/worker.py
@@ -842,13 +842,16 @@ class TaskWorkerProcessingPool:
 
         bounded_busy = max(0, min(busy, self._concurrency))
         occupancy = bounded_busy / self._concurrency if self._concurrency else 0.0
-        self._metrics.gauge(
-            "taskworker.worker.occupancy",
-            occupancy,
-            tags=tags,
-        )
-        if self._prom is not None:
-            self._prom.occupancy.labels(processing_pool=self._processing_pool_name).set(occupancy)
+        if state_counts["running"] > 0:
+            self._metrics.gauge(
+                "taskworker.worker.occupancy",
+                occupancy,
+                tags=tags,
+            )
+            if self._prom is not None:
+                self._prom.occupancy.labels(processing_pool=self._processing_pool_name).set(
+                    occupancy
+                )
 
         # Emit number of children in each state
         for state, count in state_counts.items():
--- a/clients/python/src/taskbroker_client/worker/worker.py
+++ b/clients/python/src/taskbroker_client/worker/worker.py
@@ -156,7 +156,13 @@ class TrackedChild:
     process: BaseProcess
     state: ChildState
     release: Event
-    busy: bool = False
+    # Time-weighted busy tracking. `busy_since` is the monotonic timestamp of the
+    # currently-open busy segment (None while idle); `busy_accumulated` is the busy
+    # seconds banked since the last occupancy flush. Together they let us report the
+    # fraction of the interval a child spent executing, rather than a single
+    # instantaneous busy/idle sample.
+    busy_since: float | None = None
+    busy_accumulated: float = 0.0
 
 
 class PushTaskWorker:
@@ -787,6 +793,8 @@ class TaskWorkerProcessingPool:
         self._children: Dict[UUID, TrackedChild] = {}
         self._exiting_children: Deque[UUID] = deque()
         self._children_lock = threading.Lock()
+        # Start of the interval currently being accumulated for occupancy.
+        self._last_occupancy_flush_at = time.monotonic()
         self._shutdown_event = self._mp_context.Event()
         self._prometheus_port = prometheus_port
         self._prom: WorkerPrometheusMetrics | None = None
@@ -826,7 +834,7 @@ class TaskWorkerProcessingPool:
                 extra={"error": e, "processing_pool": self._processing_pool_name},
             )
 
-        # Count the number of children in each state and waiting for exit
+        now = time.monotonic()
         with self._children_lock:
             state_counts: dict[ChildState, int] = {
                 "pending": 0,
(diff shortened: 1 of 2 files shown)


clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -826,29 +829,49 @@ def _emit_periodic_metrics(self) -> None:
...
 
-        bounded_busy = max(0, min(busy, self._concurrency))
-        occupancy = bounded_busy / self._concurrency if self._concurrency else 0.0
+        elapsed = now - self._last_occupancy_flush_at
+        self._last_occupancy_flush_at = now
+
+        running_count = state_counts["running"]
+        if running_count > 0 and elapsed > 0:
+            occupancy = busy_time / (elapsed * running_count)
+            occupancy = max(0.0, min(occupancy, 1.0))
```
george-sentry: This line implies that it's possible for `occupancy` to be negative, which would mean...

* `busy_time` is negative, or
* `elapsed` is negative, or
* `running_count` is negative, or
* All three are negative

Is that actually possible? If it isn't, it could be a good idea to simplify this line because it took me a few seconds to process.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -826,29 +829,49 @@ def _emit_periodic_metrics(self) -> None:
...
+            if self._prom is not None:
+                self._prom.occupancy.labels(processing_pool=self._processing_pool_name).set(
+                    occupancy
+                )
+
         self._metrics.gauge(
-            "taskworker.worker.occupancy",
-            occupancy,
+            "taskworker.worker.concurrency",
+            float(self._concurrency),
```
george-sentry, line 900: Is there actually a point in emitting this if it never changes? Unless it changes somewhere and I forgot.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -1017,13 +1040,19 @@ def spawn_children_thread() -> None:
                         elif message.event == "exiting":
                             self._exiting_children.append(message.child_id)
 
-                        # This child is executing a task
+                        # This child started executing a task: open a busy segment.
+                        # Guard against duplicate "busy" so we don't lose the
```
george-sentry: In what scenario would there be duplicate "busy" messages?

george-sentry: Similarly, in what scenario would there be a stray / duplicate "idle" message?

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -826,29 +829,49 @@ def _emit_periodic_metrics(self) -> None:
...
+            busy_time = 0.0
             for child in self._children.values():
                 state_counts[child.state] += 1
 
-            busy = sum(1 for child in self._children.values() if child.busy)
+                if child.busy_since is not None:
+                    child.busy_accumulated += now - child.busy_since
+                    child.busy_since = now
+                busy_time += child.busy_accumulated
+                child.busy_accumulated = 0.0
```
george-sentry: It was a bit difficult to parse all of this time calculation logic as it's intertwined with other code. Do you think it would be possible to somehow isolate it / abstract some of it away? Or would you suggest saving that for the future?

george-sentry: At least it was difficult for me 😅

george-sentry: There are specifically three different locations with very similar busy time / busy since calculations.

* Here (lines 845-849)
* When a busy message is received (lines 1047-1048)
* When an idle message is received (lines 1053 -1055)

Maybe the child could have methods for that?

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -1017,13 +1040,19 @@ def spawn_children_thread() -> None:
...
                             self._exiting_children.append(message.child_id)
 
-                        # This child is executing a task
+                        # This child started executing a task: open a busy segment.
+                        # Guard against duplicate "busy" so we don't lose the
+                        # original start time.
                         elif message.event == "busy":
-                            child.busy = True
+                            if child.busy_since is None:
+                                child.busy_since = time.monotonic()
```
fpacifici: Who is sending this message ?

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: It shouldn't be possible in theory, happy to remove the guard to simplify

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: You're right this is a per-process constant. I think it's a nice to have on our dashboards as it differs across pools and allows us to quickly compare total desired threads vs those that are warming, busy, or exiting

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: Given the current protocol, the busy and idle messages should strictly alternate per child over the shared queue. It's a cheap guard that can defend against changes to that protocol

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: Good idea, I've moved the busy logic into the `TrackedChild` so the main metric loop can simply just call those methods. Hopefully, that clears up some confusion.

clients/python/src/taskbroker_client/worker/worker.py
enochtangg: The child process's task loop sends the message. It sends busy before execution and idle in the `finally` clause. There is one multiprocessing.Queue per child drained by the parent's spawn/collect thread.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -826,29 +863,44 @@ def _emit_periodic_metrics(self) -> None:
...
 
-        bounded_busy = max(0, min(busy, self._concurrency))
-        occupancy = bounded_busy / self._concurrency if self._concurrency else 0.0
+        elapsed = now - self._last_occupancy_flush_at
+        self._last_occupancy_flush_at = now
+
+        running_count = state_counts["running"]
+        if running_count > 0 and elapsed > 0:
+            occupancy = busy_time / (elapsed * running_count)
+            occupancy = min(occupancy, 1.0)
```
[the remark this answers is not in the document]
enochtangg: Changed to read the clock after taking the lock

george-sentry (approved):

PR #779, opened 2026-08-12
fix(worker): keep pull worker healthy while queue is full

aminvakil:
We have seen failed CI runs on [getsentry/self-hosted](https://github.com/getsentry/self-hosted) many times like this:
https://github.com/getsentry/self-hosted/actions/runs/31588205812/job/94086843852?pr=4462

Some have been related to `taskbroker` being unhealthy, the reason is `taskbroker` do not mark the service healthy when its queue is full, even though worker is ready and receive tasks, when it's get long, `taskbroker` becomes unhealthy, hence the flaky CI jobs in self-hosted, this may be the intended behaviour though, I'm not sure.

This PR changes the behaviour and updates healthcheck file whenever `worker` tries to receive a task even though queue is full.

### DIFF
--- a/clients/python/src/taskbroker_client/worker/client.py
+++ b/clients/python/src/taskbroker_client/worker/client.py
@@ -328,6 +328,9 @@ class TaskbrokerClient:
             )
             self._timestamp_since_touch = cur_time
 
+    def emit_health_check(self) -> None:
+        self._emit_health_check()
+
     def _connect_to_host(self, host: str) -> ConsumerServiceStub:
         logger.info("taskworker.client.connect", extra={"host": host})
         channel = grpc.insecure_channel(host, options=self._grpc_options)
--- a/clients/python/src/taskbroker_client/worker/client.py
+++ b/clients/python/src/taskbroker_client/worker/client.py
@@ -309,7 +309,7 @@ class TaskbrokerClient:
         self._timestamp_since_touch_lock = threading.Lock()
         self._timestamp_since_touch = 0.0
 
-    def _emit_health_check(self) -> None:
+    def emit_health_check(self) -> None:
         if self._health_check_settings is None:
             return
 
@@ -327,9 +327,6 @@ class TaskbrokerClient:
                 tags={"processing_pool": self._processing_pool_name},
             )
             self._timestamp_since_touch = cur_time
-
-    def emit_health_check(self) -> None:
-        self._emit_health_check()
 
     def _connect_to_host(self, host: str) -> ConsumerServiceStub:
         logger.info("taskworker.client.connect", extra={"host": host})
@@ -406,7 +403,7 @@ class TaskbrokerClient:
         If a namespace is provided, only tasks for that namespace will be fetched.
         This will return None if there are no tasks to fetch.
         """
-        self._emit_health_check()
+        self.emit_health_check()
 
         request = GetTaskRequest(application=self._application, namespace=namespace)
         try:
@@ -458,7 +455,7 @@ class TaskbrokerClient:
 
         The return value is the next task that should be executed.
         """
-        self._emit_health_check()
+        self.emit_health_check()
         if fetch_next_task is not None:
             fetch_next_task.application = self._application
 
--- a/clients/python/src/taskbroker_client/worker/worker.py
+++ b/clients/python/src/taskbroker_client/worker/worker.py
@@ -656,6 +656,7 @@ class TaskWorker:
 
     def run_once(self) -> None:
         """Access point for tests to run a single worker loop"""
+        self.client.emit_health_check()
         self._add_task()
 
     def _add_task(self) -> bool:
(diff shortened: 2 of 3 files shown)


aminvakil:
cc @aldy505

evanh:
> Some have been related to taskbroker being unhealthy, the reason is taskbroker do not mark the service healthy when its queue is full, even though worker is ready and receive tasks, when it's get long, taskbroker becomes unhealthy, hence the flaky CI jobs in self-hosted, this may be the intended behaviour though, I'm not sure.

The health check also fires [whenever a task finishes](https://github.com/getsentry/taskbroker/blob/main/clients/python/src/taskbroker_client/worker/client.py#L458), so I'm not sure this change is addressing the actual problem. Can you expand further on the behaviour that is causing the failures?

aminvakil:
> The health check also fires [whenever a task finishes](https://github.com/getsentry/taskbroker/blob/main/clients/python/src/taskbroker_client/worker/client.py#L458), so I'm not sure this change is addressing the actual problem. Can you expand further on the behaviour that is causing the failures?

Honestly actual problem is GHA is too slow sometimes.

https://github.com/getsentry/self-hosted/actions/runs/31588205812/job/94086843852?pr=4462#step:4:5986
```
    taskworker-1                                                 | 10:45:31 [INFO] taskbroker_client.worker.worker: Running in PULL mode
```
https://github.com/getsentry/self-hosted/actions/runs/31588205812/job/94086843852?pr=4462#step:4:6016
```
    taskworker-1                                                 | 10:50:22 [INFO] sentry.hybridcloud.tasks.backfill_outboxes: backfill_outboxes.start (remaining=10000 scheduled=0)
```
In the run I've sent, it took 5 minutes for child task activity log to get shown, and they don't finish the tasks quick enough, so health file does not get recreated and container gets unhealthy, hence CI failure.

Aside from this, it's not totally unreasonable to recreate health file when parent is working but queue is full IMHO, I may be totally wrong about this though.

aminvakil:
I've also opened https://github.com/getsentry/self-hosted/pull/4464 to increase `kafka_session_timeout_ms` in `self-hosted` as another attempt to let `taskbroker` have more time when lots of containers are getting created at services startup.

evanh (approved):
> Aside from this, it's not totally unreasonable to recreate health file when parent is working, but queue is full IMHO, I may be totally wrong about this though.

Fair point.

aminvakil:
Failed CI is irrelevant to this PR.
```
  #21 ERROR: failed to push ghcr.io/getsentry/taskbroker-python-example:d7a8d7e7d88433686927fd661375e6f558ea1381: denied: installation not allowed to Write organization package
```

aminvakil:
I opened two PRs to fix `python-example-image` workflow (https://github.com/getsentry/taskbroker/pull/781, https://github.com/getsentry/taskbroker/pull/782).

This can wait after those get merged, because without #781 , we can not get green CI in this PR.

aminvakil:
> I opened two PRs to fix `python-example-image` workflow (#781, #782).
> 
> This can wait after those get merged, because without #781 , we can not get green CI in this PR.

They got merged and I rebased this branch on top of main and pushed, hopefully CI will pass now.

clients/python/src/taskbroker_client/worker/client.py
```diff
@@ -328,6 +328,9 @@ def _emit_health_check(self) -> None:
             )
             self._timestamp_since_touch = cur_time
 
+    def emit_health_check(self) -> None:
+        self._emit_health_check()
```
markstory: Why not rename the existing method? This feels like unnecessary indirection.

clients/python/src/taskbroker_client/worker/worker.py
```diff
@@ -656,6 +656,7 @@ def signal_handler(*args: Any) -> None:
 
     def run_once(self) -> None:
         """Access point for tests to run a single worker loop"""
+        self.client.emit_health_check()
```
markstory: If this is happening on every run loop iteration, does it also need to happen in each `get_task()` and `update_task()` call?  Emitting the health check here would make those calls redundant.

clients/python/src/taskbroker_client/worker/client.py
aminvakil: Because I (my AI agent) wanted to keep changes minimum and do not change current methods, that was naive, I'll change that.

clients/python/src/taskbroker_client/worker/worker.py
aminvakil: I will move `health_check` emitting from every `run_once` to `self.worker_pool.is_worker_full`, so the change would be exactly what PR title says and addresses the problem we're going to fix here.

clients/python/src/taskbroker_client/worker/client.py
aminvakil: Fixed in https://github.com/getsentry/taskbroker/pull/779/commits/80815f5dc7d7e8947ddeab178e8c71d3df5d1d74.

clients/python/src/taskbroker_client/worker/worker.py
aminvakil: Fixed in https://github.com/getsentry/taskbroker/pull/779/commits/80815f5dc7d7e8947ddeab178e8c71d3df5d1d74.

PR #794, opened 2026-09-17
fix(fetch): Undo the claim when submitting to the push pool fails

enochtangg:
## Description

When the fetch thread claims an activation but cannot hand it to the push pool, the row is left in the `Claimed` state with nothing holding it. The claim query only selects `Pending` rows, so nothing re-picks it and it sits until `handle_claim_expiration` reverts it.

This inflates the SLO metric. `pending_activation.max_lag.sec` counts `Claimed` rows, so one stranded activation drives it up until the lease expires. This symptom can be observed consistently in `process-segments-push` in s4s2: [DD link](https://app.datadoghq.com/dashboard/zn3-zme-x9x/taskbroker-push-alloydb?fromUser=true&fullscreen_end_ts=1789635654362&fullscreen_paused=true&fullscreen_refresh_mode=paused&fullscreen_section=overview&fullscreen_start_ts=1789634694042&fullscreen_widget=1734857525129951&refresh_mode=paused&tpl_var_alloydb_instance%5B0%5D=db-taskbroker-default-primary&tpl_var_broker_replica%5B0%5D=%2A&tpl_var_horizontalpodautoscaler%5B0%5D=%2A&tpl_var_kafka_broker_cluster%5B0%5D=%2A&tpl_var_kafka_consumer_group%5B0%5D=taskbroker-process-segments-push&tpl_var_kafka_topic%5B0%5D=taskworker-process-segments-push&tpl_var_kube_name%5B0%5D=task-process-segments-push-worker%2A&tpl_var_kube_name%5B1%5D=task-process-segments-push-broker%2A&tpl_var_processing_pool%5B0%5D=process-segments-push&tpl_var_sentry_region%5B0%5D=s4s2&tpl_var_worker_kube_name%5B0%5D=%2A&from_ts=1789634694042&to_ts=1789635654362&live=false). During this time, throughput, occupancy and AlloyDB latency were all normal, so the metric was tracking a single stuck row rather than real pipeline latency.

When workers briefly  go unavailable (due to deployment), pushes fail each activation revert its own claim which is correct, but that work filled the push queue to its cap. Activations that hit the push queue and receive a submit timeout don't get reverted and only log

## Fix

Undo the claim on both `QueueError` arms. This should be safe because the activation never left the broker, so no worker can be running it and reverting cannot cause a double execution. A submit failure now costs roughly the fetch backoff instead of the full claim lease.

The push thread already did this inline after a failed gRPC push, so the shared logic moved to `ActivationStore::undo_claim`.


src/store/traits.rs
```diff
@@ -97,6 +97,29 @@ pub trait ActivationStore: Send + Sync {
         delay_on_retry: Option<u64>,
     ) -> Result<Option<Activation>, Error>;
 
+    /// Release a claim on an activation, returning it to pending.
+    ///
+    /// Both the fetch and push pools need this when an activation was claimed
+    /// but could not be handed to a worker.
+    async fn undo_claim(&self, id: &str, metric: &'static str) {
+        if let Err(e) = self
+            .set_status(id, ActivationStatus::Pending, None, None)
```
evanh: Maybe we need to be more defensive here, and only update the status if it is in the Claimed state (`UPDATE ... SET status = 'Pending' WHERE id = X AND status = 'Claimed'). That way we avoid any race conditions (e.g. https://github.com/getsentry/taskbroker/pull/794/changes#r4041591106).

enochtangg: Good idea, updated such that release_claim only updates the state if the task was in claimed state

src/fetch/thread.rs
```diff
@@ -125,6 +125,9 @@ impl FetchThread {
                                 self.config.push.queue.timeout.as_millis()
                             );
 
+                            // Nothing else will push this activation, so release the claim
+                            self.store.undo_claim(&id, "fetch.undo_claim").await;
+
```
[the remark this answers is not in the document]
enochtangg: Switched to try_send in a retry loop, so a failed send hands the activation back and we know the push pool never got it. This removes the racing an undo against a delivery that already landed.

src/fetch/thread.rs
```diff
@@ -137,6 +142,9 @@ impl FetchThread {
                                 "Submit to push pool failed due to closed channel",
                             );
 
+                            // Nothing else will push this activation, so release the claim
+                            self.store.undo_claim(&id, "fetch.undo_claim").await;
+
                             // We cannot recover from a closed channel
                             return false;
```
untitaker, line 149: if we abort early here, then the remaining ids in the batch are still stuck in Claimed, right?

also isn't this fundamentally racy? if the claim expires before we get to this line, we might undo somebody else's claim.

untitaker: just tested it and fable was able to find the same thing (and more things i didn't verify)

evanh: As long as the timeout for `QueueError::Timeout` is shorter than the claim duration, we shouldn't have a race here. This is happening in the fetch thread, so it's not possible for the status of these tasks to change without them being successfully pushed to the queue first.

The race condition here would be if the claim duration < queue timeout. Then the race would be:
- Activation is fetched and marked as CLAIMED
- Fetcher blocks trying to write to the queue
- claim expires
- Upkeep moves activation from CLAIMED to PENDING
- A different fetcher thread fetches the activation and updates to CLAIMED
- Original fetcher queue times out, reverts activation back to PENDING

